PR #17, opened 2026-06-23
Fix export segment const lint

psiddharthdesign:
## Summary
- Change the export frame segment lower bound from `let` to `const` because it is never reassigned.
- Keep the upper bound mutable because it is clamped when needed.

## Verification
- `pnpm exec eslint src/export/formats.ts`

## Notes
- Fresh baseline repo-wide `pnpm lint` and `pnpm typecheck` currently fail on unrelated existing errors across multiple files, so this PR verifies the touched file only.


psiddharthdesign:
@mergifyio queue

PR #28, opened 2026-06-24
Tighten CLI scene default types

psiddharthdesign:
## Summary
- add explicit local types for CLI scene transform and size defaults
- remove casts when merging agent-supplied transform and size patches

## Verification
- pnpm --dir cli test
- git diff --check

Note: root `pnpm typecheck` and `pnpm lint` are currently red on unrelated pre-existing app errors outside this CLI change.


psiddharthdesign:
@mergifyio queue

PR #33, opened 2026-06-24
Preserve declared node ids in CLI scenes

psiddharthdesign:
## Summary\n- key CLI-authored scene nodes by each node's declared id\n- infer root and active camera from declared ids when not provided\n- add regression coverage for arbitrary node map keys\n\n## Tests\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #36, opened 2026-06-24
fix(cli): count Y.Array keyframes in scene summary

psiddharthdesign:
## Summary
- count scene summary keyframes when a track stores keyframes as a Y.Array
- add a CLI regression test for the alternate persisted representation

## Verification
- pnpm test (in cli/)

Note: root `pnpm typecheck` was attempted but currently fails on pre-existing app-wide TypeScript errors unrelated to this CLI change.


psiddharthdesign:
@mergifyio queue

PR #38, opened 2026-06-24
fix(cli): key authored tracks by declared id

psiddharthdesign:
## Summary
- store CLI-authored tracks under their declared track id instead of the surrounding object key
- add regression coverage matching the existing node-id behavior

## Tests
- cd cli && pnpm test


psiddharthdesign:
@mergifyio queue

PR #39, opened 2026-06-24
fix(cli): key authored sections by declared id

psiddharthdesign:
## Summary
- Store authored scene sections under their declared section id, matching node and track serialization behavior.
- Add CLI scene builder coverage for aliased section map keys.

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #48, opened 2026-06-24
Clarify MCP render scene handling

psiddharthdesign:
## Summary
- Clarify that the MCP render_scene tool currently renders the desktop scene even when a scene path is provided
- Include an explicit note in the tool response when a provided scene path is ignored

## Verification
- pnpm --dir cli test
- pnpm --dir cli run build

Note: root pnpm run typecheck and pnpm run lint were attempted, but fail on existing unrelated repository issues outside this change.


psiddharthdesign:
@mergifyio queue

PR #50, opened 2026-06-24
Clarify CLI scene ID handling

psiddharthdesign:
## Summary\n- Clarifies that CLI scene builder map keys come from declared node/track/section IDs, while input record keys may be aliases.\n\n## Tests\n- pnpm test (in cli/)


psiddharthdesign:
@mergifyio queue

PR #51, opened 2026-06-24
Preserve image import warnings in CLI scenes

psiddharthdesign:
## Summary\n- add the optional image importWarning field to the CLI scene schema\n- persist import warnings when creating .hype files from JSON\n- cover the round-trip in the scene-builder tests\n\n## Tests\n- pnpm test


psiddharthdesign:
@mergifyio queue

PR #54, opened 2026-06-24
Fix CLI camera projection type

psiddharthdesign:
## Summary
- Align the CLI scene JSON camera projection type with the desktop scene type by accepting `perspective`
- Add a CLI builder test that verifies perspective camera projection round-trips into .hype bytes

## Tests
- `pnpm --dir cli test`


psiddharthdesign:
@mergifyio queue

PR #58, opened 2026-06-24
chore(cli): name keyframe json type

psiddharthdesign:
## Summary
- extract the CLI scene keyframe schema into an exported KeyframeJson type
- keep TrackJson behavior unchanged while making the authoring schema easier to reference

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #59, opened 2026-06-24
Tighten CLI layout guide scene type

psiddharthdesign:
## Summary
- add a concrete LayoutGuideJson type for CLI scene layout guide metadata
- type NodeJson.layoutGuides with that schema instead of unknown[]

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #62, opened 2026-06-25
Clone CLI scene defaults deeply

psiddharthdesign:
## Summary
- Deep-copy CLI scene default objects before applying partial patches
- Add a regression test proving sibling frame layout padding defaults stay independent

## Verification
- pnpm test (in cli/)

Note: root pnpm typecheck currently fails on existing app-level TypeScript errors outside this CLI change.


psiddharthdesign:
@Mergifyio queue

PR #71, opened 2026-06-25
Preserve tracks in scene snapshots

psiddharthdesign:
## Summary
- add a read-only SceneAPI helper for all animation tracks
- include tracks when snapshotting a scene so exported scene objects preserve authored animations

## Verification
- PASS: pnpm exec eslint src/scene/doc.ts
- PASS: git diff --check
- FULL CHECK NOT CLEAN: pnpm lint and pnpm typecheck currently fail on pre-existing issues outside src/scene/doc.ts (for example React hook lint errors in RenderWindowApp/Canvas and existing TypeScript errors around SceneNode exports and NodeBaseMutable keys).

### DIFF
--- a/src/scene/doc.ts
+++ b/src/scene/doc.ts
@@ -95,6 +95,8 @@ export interface SceneAPI {
   /** All node ids in the scene. */
   getAllNodeIds(): NodeId[]
   getTrack(id: TrackId): Track | null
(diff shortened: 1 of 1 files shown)


psiddharthdesign:
@mergifyio queue

PR #77, opened 2026-06-25
test(cli): cover text node defaults

psiddharthdesign:
## Summary
- add CLI scene-builder assertions for text node default fields
- keep the change limited to test coverage for existing behavior

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #101, opened 2026-06-25
Clarify CLI scene render docs

psiddharthdesign:
## Summary
- Clarify that `hypermotion render --scene` is accepted for forward compatibility but still renders the current desktop scene today.
- Keep the CLI quick-reference aligned with the current render command behavior.

## Tests
- `pnpm --dir cli test`


psiddharthdesign:
@mergifyio queue

PR #103, opened 2026-06-25
docs(cli): clarify render scene path option

psiddharthdesign:
## Summary
- update the CLI render request comment for scenePath now that file-based scene rendering is wired through

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #104, opened 2026-06-25
test(cli): cover camera point of interest defaults

psiddharthdesign:
## Summary
- add CLI scene-builder coverage for camera point-of-interest defaults derived from transform coordinates

## Verification
- pnpm --dir cli test

## Notes
- Broader root `pnpm typecheck` currently fails on pre-existing app-wide TypeScript errors unrelated to this test-only change.


psiddharthdesign:
@mergifyio queue

PR #107, opened 2026-06-25
chore(cli): name scene node kind type

psiddharthdesign:
## Summary
- name the CLI scene node kind union as NodeKindJson
- reuse the type for node schema helpers

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #110, opened 2026-06-25
Add CLI support for anchor keyframe properties

psiddharthdesign:
## Summary
- add transform anchor property ids to the CLI scene schema
- cover anchor keyframe preservation in the CLI scene builder tests

## Verification
- pnpm --dir cli test

Note: root `pnpm typecheck` was also attempted, but it currently fails on existing app-wide TypeScript errors unrelated to this CLI change.


psiddharthdesign:
@mergifyio queue

PR #111, opened 2026-06-25
chore(cli): type render command options

psiddharthdesign:
## Summary
- add an explicit local type for CLI render command options
- keep render option handling aligned with the known Commander fields

## Verification
- pnpm --dir cli test

Note: root `pnpm run typecheck` currently fails on unrelated existing main-branch TypeScript errors, so verification used the focused CLI package test suite.


psiddharthdesign:
@mergifyio queue

PR #113, opened 2026-06-25
Type CLI scene appearance input

psiddharthdesign:
## Summary
- Add an explicit CLI scene appearance input type with per-corner radii support
- Add a focused CLI authoring test that verifies per-corner radii are preserved in .hype output

## Verification
- PASS: pnpm --dir cli test
- NOTE: pnpm typecheck currently fails on existing app-wide TypeScript issues outside this change
- NOTE: pnpm lint currently fails on existing app-wide lint issues outside this change


psiddharthdesign:
@mergifyio queue

PR #114, opened 2026-06-25
Clarify CLI create frame example

psiddharthdesign:
## Summary
- Update the `hypermotion create` command comment example so the sample frame includes the required `size` and `layout` fields.

## Verification
- `pnpm test` in `cli/`


psiddharthdesign:
@mergifyio queue

PR #115, opened 2026-06-25
Clarify terminal render workflow

psiddharthdesign:
## Summary
- update AGENTS.md terminal workflow to use the supported current-scene render command
- note that passing a .hype scene path through render is currently MCP-only

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #117, opened 2026-06-25
Fill default media size axes in CLI scenes

psiddharthdesign:
## Summary
- default missing media size axes when building CLI-authored scenes
- add a regression test for partial video size payloads

## Verification
- cd cli && pnpm test


psiddharthdesign:
@mergifyio queue

PR #119, opened 2026-06-25
Allow JSON variant keyframe values

psiddharthdesign:
## Summary
- Widen CLI keyframe value typing so variant keyframes can carry JSON object fields beyond strings
- Add a regression test covering boolean and numeric variant keyframe fields

## Verification
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #121, opened 2026-06-25
Allow partial CLI scene sizes

psiddharthdesign:
## Summary
- add a public SizeJson type that permits omitted width or height axes
- remove the test cast now that partial media sizes typecheck directly

## Verification
- pnpm test (in cli/)


psiddharthdesign:
@mergifyio queue

PR #122, opened 2026-06-25
docs: include camera transform in CLI create example

psiddharthdesign:
## Summary
- update the CLI create command's inline JSON example so the camera node includes the required transform fields

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #129, opened 2026-06-25
Allow null CLI keyframe values

psiddharthdesign:
## Summary
- Allow CLI-authored keyframes to use null values, matching properties such as appearance.fill
- Add a focused scene-builder round-trip test for null keyframe values

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #130, opened 2026-06-25
Allow per-corner radius scene tracks

psiddharthdesign:
## Summary
- add per-corner radius property IDs to the CLI scene authoring schema
- cover per-corner radius keyframe round-tripping in the scene builder tests

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #134, opened 2026-06-25
Add CLI media fit type alias

psiddharthdesign:
## Summary
- Add a shared exported MediaFitJson alias for CLI image/video node fit values.
- Use the alias from NodeJson so the CLI scene schema has one named media-fit surface.

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #136, opened 2026-06-25
Clarify agent render workflow

psiddharthdesign:
## Summary
- Clarify that agent-created .hype files can be inspected with info_scene, but render_scene still renders the current desktop scene.
- Note that render_scene currently ignores .hype scene paths until file-based headless rendering lands.

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #137, opened 2026-06-25
chore(cli): type variant transitions

psiddharthdesign:
## Summary
- add a typed CLI scene schema shape for component variant transitions
- keep the existing default transition behavior unchanged

## Tests
- pnpm --dir cli test

## Notes
- Root `pnpm run typecheck` and `pnpm run lint` currently fail on pre-existing app-wide issues outside this CLI-only change.


psiddharthdesign:
@mergifyio queue

PR #146, opened 2026-06-25
Clarify MCP create_scene follow-up text

psiddharthdesign:
## Summary
- Clarifies the create_scene MCP success text so it tells users to open the generated .hype file in the desktop app for hand-editing or rendering.
- Avoids implying render_scene directly renders the newly created file when the current headless path still renders the desktop scene.

## Verification
- pnpm test (from cli/)


psiddharthdesign:
@mergifyio queue

PR #147, opened 2026-06-25
test(cli): clean up info command temp dir

psiddharthdesign:
## Summary
- clean up the temporary directory created by the CLI info command test
- keep stdout restoration inside its existing finally block

## Verification
- pnpm --dir cli test

## Notes
- Attempted root `pnpm typecheck && pnpm lint`, but root typecheck currently fails on unrelated existing app/export/UI TypeScript errors before lint runs.


psiddharthdesign:
@mergifyio queue

PR #151, opened 2026-06-25
Deep merge scene metadata defaults

psiddharthdesign:
## Summary
- deep-merge scene metadata defaults in the CLI builder so partial nested metadata keeps default canvas fields
- add a focused regression test for partial canvas metadata

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #154, opened 2026-06-25
test(cli): cover nested create output dirs

psiddharthdesign:
## Summary
- add CLI create command coverage for recursively creating nested output directories
- verify the written .hype file can be read back with the expected scene metadata

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #165, opened 2026-06-25
Test explicit audio timing serialization

psiddharthdesign:
## Summary
- add CLI scene serialization coverage for explicit audio timing fields
- verify audio duration, volume, trim, loop, and muted values round-trip into .hype bytes

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #172, opened 2026-06-25
test: assert default node visibility in scene builder

psiddharthdesign:
## Summary
- add a focused CLI scene builder assertion for the default node visibility persisted into .hype files

## Verification
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #174, opened 2026-06-25
Tighten CLI stdout test helper types

psiddharthdesign:
## Summary\n- use Node's actual stdout.write parameter types in the CLI stdout capture test helper\n- keep the helper behavior unchanged while avoiding a stale hand-written stream signature\n\n## Verification\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #181, opened 2026-06-25
test: cover info output without camera

psiddharthdesign:
## Summary\n- add CLI info command coverage for scenes without an active camera\n- verify the human-readable summary omits the camera line in that case\n\n## Tests\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #189, opened 2026-06-26
Allow clearing CLI active camera

psiddharthdesign:
## Summary\n- allow CLI-authored scenes to explicitly clear activeCameraId with null\n- add a regression test so null is preserved instead of inferring the first camera\n\n## Tests\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #190, opened 2026-06-26
Clarify CLI render scene behavior

psiddharthdesign:
## Summary
- clarify that the CLI render command always uses the current desktop scene today
- remove misleading quick-reference wording around passing `--scene` as a usable workflow

## Verification
- `pnpm test` in `cli/`


psiddharthdesign:
@mergifyio queue

PR #193, opened 2026-06-26
docs(cli): clarify render scene flag status

psiddharthdesign:
## Summary
- Update the CLI render command header comment to match the current `--scene` behavior.
- Clarify that the flag is accepted for compatibility but current desktop scene rendering is still used until file-based headless rendering lands.

## Verification
- `pnpm test` in `cli/`


psiddharthdesign:
@mergifyio queue

PR #198, opened 2026-06-26
Add named options type for info command

psiddharthdesign:
## Summary
- add a named local options type for the CLI info command
- keep Commander action typing explicit without changing behavior

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #208, opened 2026-06-26
Clarify render_scene scene path behavior

psiddharthdesign:
## Summary
- Clarify MCP render_scene docs/schema text so agents know scene paths are accepted for forward compatibility but current renders still target the desktop scene.

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #213, opened 2026-06-26
Guard create_scene transform property docs

psiddharthdesign:
## Summary\n- add a focused CLI MCP test that ensures create_scene documents supported transform property IDs\n- helps catch future drift between the agent-facing authoring docs and the typed PropertyId surface\n\n## Verification\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #229, opened 2026-06-26
Add locator test for missing app path override

psiddharthdesign:
## Summary
- add CLI locator coverage for a missing HYPERMOTION_APP_PATH override
- verify the locator returns null and surfaces the intended diagnostic without leaking env changes

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #230, opened 2026-06-26
Share CLI keyframe property ids

psiddharthdesign:
## Summary
- export the CLI keyframe property id list from the scene schema module
- reuse that list in MCP capabilities responses to avoid a duplicate contract

## Verification
- pnpm --dir cli test

Note: top-level pnpm lint/typecheck currently fail on unrelated existing app issues from origin/main; this PR is CLI-only and was verified with the CLI package suite.


psiddharthdesign:
@mergifyio queue

PR #233, opened 2026-06-26
Allow CLI patches to clear active camera

psiddharthdesign:
## Summary
- widen the CLI scene patch type so setActiveCameraId can clear the camera with null
- add regression coverage for clearing activeCameraId through applyScenePatch

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #237, opened 2026-06-26
test: cover full MCP keyframe property list

psiddharthdesign:
## Summary
- Assert MCP capability tools return the full exported keyframe property list instead of a handpicked subset.
- Keeps capability tests tied to the scene builder source of truth to catch future drift.

## Verification
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #238, opened 2026-06-26
fix(cli): default workspaceOnly in scene patches

psiddharthdesign:
## Summary\n- default workspaceOnly to false for nodes created via applyScenePatch\n- add regression coverage for patch-created node defaults\n\n## Tests\n- pnpm test (in cli/)


psiddharthdesign:
@mergifyio queue

PR #240, opened 2026-06-26
Tighten error boundary catch info type

psiddharthdesign:
## Summary
- type React error boundary catch info as ErrorInfo instead of unknown
- remove a stale eslint-disable directive from the same handler

## Verification
- pnpm exec eslint src/ui/ErrorBoundary.tsx
- pnpm exec tsc --ignoreConfig --noEmit --strict --jsx react-jsx --target es2023 --lib ES2023,DOM,DOM.Iterable --module esnext --moduleResolution bundler --verbatimModuleSyntax --skipLibCheck src/ui/ErrorBoundary.tsx

Note: full repo checks were attempted after installing dependencies from the existing lockfile, but current main has unrelated pre-existing typecheck/lint failures outside this touched file.

### DIFF
--- a/src/ui/ErrorBoundary.tsx
+++ b/src/ui/ErrorBoundary.tsx
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
-import { Component, type ReactNode } from 'react'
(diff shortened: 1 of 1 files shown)


psiddharthdesign:
@mergifyio queue

PR #241, opened 2026-06-26
Validate track property ids

psiddharthdesign:
## Summary
- reject unsupported track property ids during CLI scene validation
- cover the validation path with a focused scene unit test

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #242, opened 2026-06-26
Validate scene root node kind

psiddharthdesign:
## Summary\n- reject .hype scenes whose scene.root points at a non-frame node\n- add CLI scene-builder coverage for the validation error\n\n## Tests\n- pnpm test


psiddharthdesign:
@mergifyio queue

PR #245, opened 2026-06-26
fix(cli): default patched text nodes

psiddharthdesign:
## Summary
- apply text-node defaults when CLI patch createNode authors a text layer
- add regression coverage for partial text size defaulting

## Tests
- pnpm test (cli)


psiddharthdesign:
@mergifyio queue

PR #246, opened 2026-06-26
Stabilize locator path assertion

psiddharthdesign:
## Summary\n- avoid constructing a regex from a temporary filesystem path in the locator test\n- assert the captured error message contains the path literally instead\n\n## Tests\n- pnpm test


psiddharthdesign:
@mergifyio queue

PR #252, opened 2026-06-26
chore(cli): type scene query MCP handlers

psiddharthdesign:
## Summary
- add explicit CallToolResult return types to scene query/open MCP handlers
- type the shared MCP text response helper

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #254, opened 2026-06-26
Test per-side stroke width scene output

psiddharthdesign:
## Summary
- add CLI scene-builder coverage for preserving per-side stroke widths in serialized appearance data

## Verification
- pnpm --dir cli test -- --test-name-pattern="buildSceneBytes"
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #256, opened 2026-06-26
test(cli): cover invalid active camera validation

psiddharthdesign:
## Summary
- add CLI validate command coverage for activeCameraId pointing at a non-camera node
- assert the validator reports an error instead of the missing-camera warning

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #257, opened 2026-06-26
Validate scene node id keys

psiddharthdesign:
## Summary
- add a scene validation error when a node map key does not match the node's stored id
- cover the mismatch with a focused CLI scene validator regression test

## Verification
- PASS: pnpm --dir cli test
- BLOCKED/PRE-EXISTING: pnpm typecheck and pnpm lint trigger the root install approval flow for electron@33.4.11 in this environment; running the underlying binaries directly shows existing app-wide type/lint failures unrelated to this CLI validator change.


psiddharthdesign:
@mergifyio queue

PR #260, opened 2026-06-26
Validate inverse scene parent links

psiddharthdesign:
## Summary
- report a validation error when a node's parent does not list it in children
- add a CLI regression test for the inconsistent parent/children link

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #262, opened 2026-06-26
Tighten invalid property test typing

psiddharthdesign:
## Summary
- replace a broad unknown cast in the invalid property validation test with a targeted @ts-expect-error
- keep the runtime validation coverage for unsupported scene track property IDs

## Tests
- pnpm test (from cli/)


psiddharthdesign:
@mergifyio queue

PR #265, opened 2026-06-26
docs: refresh fill renderer comment

psiddharthdesign:
## Summary
- update the fill utility header comment now that the Pixi export renderer exists
- keep the change comment-only and scoped to scene fill serialization context

## Verification
- npx eslint src/scene/fill.ts

## Notes
- Full repo lint/typecheck were attempted but are currently blocked by existing unrelated lint/type errors on main.


psiddharthdesign:
@mergifyio queue

PR #267, opened 2026-06-26
Validate CLI track id consistency

psiddharthdesign:
## Summary
- validate saved track map keys against each track's declared id
- add focused coverage for mismatched persisted track ids

## Tests
- npm --prefix cli test


psiddharthdesign:
@mergifyio queue

PR #268, opened 2026-06-26
Normalize empty scene ids in CLI summaries

psiddharthdesign:
## Summary
- normalize empty-string scene root and active camera ids to null in CLI summaries
- add regression coverage for delete-node cleanup paths

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #269, opened 2026-06-26
Default patched shape node sizes

psiddharthdesign:
## Summary
- default size for rect, ellipse, and image nodes created through scene patches
- add regression coverage for minimal patched shape nodes

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #270, opened 2026-06-26
Tighten CLI scene summary metadata types

psiddharthdesign:
## Summary
- add exported scene summary metadata types for CLI consumers
- keep summary metadata partial to support older or hand-built .hype files
- cover typed canvas access in the existing scene builder test

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #271, opened 2026-06-26
test(cli): cover patch_scene alternate output

psiddharthdesign:
## Summary
- add MCP coverage for patch_scene writing to an alternate output path
- verify applyLive: false avoids the live app path while preserving the source scene

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #277, opened 2026-06-26
Validate render_scene scene paths

psiddharthdesign:
## Summary
- Return an MCP error when render_scene receives a missing .hype scene path
- Add coverage for the missing-scene case

## Tests
- pnpm test (cli)


psiddharthdesign:
@mergifyio queue

PR #278, opened 2026-06-26
Avoid ts-expect-error in scene validation test

psiddharthdesign:
## Summary
- Replace a test-only TypeScript suppression with an explicit unknown boundary for intentionally invalid scene input
- Keep runtime validation coverage for unsupported track property ids unchanged

## Tests
- pnpm test (cli)


psiddharthdesign:
@mergifyio queue

PR #296, opened 2026-06-26
Validate scene root has no parent

psiddharthdesign:
## Summary
- reject saved scenes where the root frame has a parent
- add focused CLI validation coverage for malformed persisted scenes

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #297, opened 2026-06-26
test(cli): cover empty app path override

psiddharthdesign:
## Summary
- add CLI locator coverage for an empty HYPERMOTION_APP_PATH
- assert the empty override behaves like an unset value without emitting the explicit-override error

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #298, opened 2026-06-26
Type patch_scene MCP handler result

psiddharthdesign:
## Summary
- Add an explicit `CallToolResult` return type to the `patch_scene` MCP handler.

## Verification
- `pnpm --dir cli test`


psiddharthdesign:
@mergifyio queue

PR #305, opened 2026-06-27
Validate scene node kinds

psiddharthdesign:
## Summary\n- add explicit supported node kind validation for .hype scene files\n- cover unsupported node kinds in the scene validator tests\n\n## Tests\n- pnpm test


psiddharthdesign:
@mergifyio queue

PR #318, opened 2026-06-27
Keep create_scene property docs in sync

psiddharthdesign:
## Summary
- add a coverage check that the create_scene MCP description lists every supported keyframe property id
- reuse the exported PROPERTY_IDS list so the test catches future schema/documentation drift

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #324, opened 2026-06-27
test(cli): cover valid scene validation

psiddharthdesign:
## Summary
- add a positive validateScene test for a valid authored scene

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #330, opened 2026-06-27
Use SDK result type for capability tools

psiddharthdesign:
## Summary\n- Replace the local MCP text result shape in capability tools with the SDK CallToolResult type\n- Update the capability test helper to assert text content before parsing\n\n## Tests\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #334, opened 2026-06-27
Add typed text animation scene config

psiddharthdesign:
## Summary
- add an explicit CLI SceneJson type for text animation track config
- annotate the existing patch-scene test fixture with that type

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #338, opened 2026-06-27
test: cover patched text animation defaults

psiddharthdesign:
## Summary\n- Add regression coverage that patched text animation tracks default omitted keyframes to an empty array.\n- Keep this as a CLI scene-builder test-only maintenance change.\n\n## Verification\n- pnpm --dir cli test\n\n## Notes\n- Repository-level pnpm typecheck/lint did not reach checks because pnpm 11 blocked Electron install scripts with ERR_PNPM_IGNORED_BUILDS.


psiddharthdesign:
@mergifyio queue

PR #343, opened 2026-06-27
Tighten render scene schema test typing

psiddharthdesign:
## Summary
- Add a small typed helper for render_scene input schema assertions
- Remove repeated loose Record casts in the render_scene MCP tests

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #344, opened 2026-06-27
Fix patch-created frame defaults

psiddharthdesign:
## Summary\n- fill default size, layout, and frame/component/instance metadata when CLI patch operations create nodes\n- add regression coverage for patch-created frame layout defaults\n\n## Tests\n- pnpm test


psiddharthdesign:
@mergifyio queue

PR #345, opened 2026-06-27
Test validate command missing file handling

psiddharthdesign:
## Summary
- add CLI validate command coverage for missing scene files
- assert the existing stderr message and exit code path

## Verification
- cd cli && ./node_modules/.bin/tsc && node --test dist/commands/validate.test.js && node --test dist/**/*.test.js

Note: root pnpm lint currently stops during dependency setup because pnpm 11 ignores package.json#pnpm.onlyBuiltDependencies and blocks Electron build scripts.


psiddharthdesign:
@mergifyio queue

PR #346, opened 2026-06-27
Stabilize doctor missing app tests

psiddharthdesign:
## Summary
- use per-test temporary missing app paths in doctor CLI and MCP tests
- clean up temporary directories after each test

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #349, opened 2026-06-27
Tighten decoded scene test record types

psiddharthdesign:
## Summary
- add named decoded-record aliases in scene build tests
- replace repeated ad hoc decoded map casts with the aliases

## Tests
- pnpm --dir cli install --frozen-lockfile --ignore-scripts
- pnpm --dir cli test

Note: root pnpm install was attempted first, but pnpm v11 stopped on the Electron build-script approval gate before tests ran. The CLI-only install/test path passed.

### DIFF
--- a/cli/src/scene/build.test.ts
+++ b/cli/src/scene/build.test.ts
@@ -12,6 +12,9 @@ import {
   type TextAnimationJson,
 } from './build.js'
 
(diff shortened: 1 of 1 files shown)


psiddharthdesign:
@mergifyio queue

PR #352, opened 2026-06-27
Widen CLI stream capture callback type

psiddharthdesign:
## Summary
- allow CLI stream capture helpers to accept sync callbacks as well as async callbacks
- keep runtime behavior unchanged by reusing the existing await path

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #356, opened 2026-06-27
Handle missing query scene files

psiddharthdesign:
## Summary\n- return structured MCP errors when query scene handlers cannot read a .hype file\n- add coverage for missing scene files across list_layers, get_layer, list_tracks, and list_cameras\n\n## Tests\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #357, opened 2026-06-27
chore: type CLI fatal handler

psiddharthdesign:
## Summary
- annotate the CLI fatal parse handler parameter as unknown
- keep the existing Error narrowing explicit before logging

## Verification
- CI=true npx pnpm@9 install --frozen-lockfile && npx pnpm@9 test (from cli/)


psiddharthdesign:
@mergifyio queue

PR #362, opened 2026-06-27
Share MCP patch operation capabilities

psiddharthdesign:
## Summary
- export the supported patch operation names from the scene builder
- reuse that typed list in MCP capabilities responses and tests

## Tests
- pnpm test (cli)


psiddharthdesign:
@mergifyio queue

PR #364, opened 2026-06-27
test(cli): cover inspect missing file errors

psiddharthdesign:
## Summary
- add CLI inspect command coverage for missing scene file errors
- assert the command exits with code 2 and reports the read failure on stderr

## Tests
- pnpm test

PR #365, opened 2026-06-27
test: cover inspect missing scene files

psiddharthdesign:
## Summary
- add a CLI regression test for `hypermotion inspect` when the scene file is missing
- assert the command reports the read failure on stderr and exits with code 2

## Tests
- `pnpm --dir cli test`


psiddharthdesign:
Closing as redundant: PR #364 already merged the same inspect missing-file coverage after this branch was created. I’ll open a fresh maintenance PR from current main.

PR #366, opened 2026-06-27
Validate scene section id consistency

psiddharthdesign:
## Summary
- validate section map keys against section ids in CLI scene validation
- add regression coverage for mismatched section ids

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #371, opened 2026-06-27
test(cli): cover malformed inspect files

psiddharthdesign:
## Summary
- add a CLI regression test for `hypermotion inspect` when a `.hype` file cannot be decoded
- assert the command reports the inspect failure on stderr and exits with code 2

## Tests
- `pnpm --dir cli test`


psiddharthdesign:
@mergifyio queue

PR #393, opened 2026-06-28
Handle blank scene names in CLI info

psiddharthdesign:
## Summary
- Treat blank scene names as unnamed in the human-readable `hypermotion info` output.
- Add a CLI regression test for whitespace-only scene names.

## Tests
- `pnpm --dir cli test`


psiddharthdesign:
@mergifyio queue

PR #400, opened 2026-06-28
test(cli): isolate doctor env overrides

psiddharthdesign:
## Summary
- add a small CLI test helper for scoped environment variable overrides
- use it in doctor command and MCP doctor tests to avoid duplicated restore logic

## Verification
- pnpm test (in cli)


psiddharthdesign:
@mergifyio queue

PR #405, opened 2026-06-28
Type CLI capability metadata

psiddharthdesign:
## Summary
- define MCP capability tool/format lists as readonly literal constants
- type the capabilities payload from those constants while preserving the JSON response shape

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #406, opened 2026-06-28
Clean up removed custom font faces

psiddharthdesign:
## Summary\n- unregister custom FontFace entries when scene custom fonts are removed\n- refresh layout metrics after removed font cleanup\n- update the hook comment now that cleanup is implemented\n\n## Verification\n- ./node_modules/.bin/eslint src/ui/fonts/useCustomFonts.ts\n- git diff --check origin/main..HEAD\n\nNote: repository-wide pnpm run typecheck/lint currently cannot run in this environment because pnpm rejects the Electron ignored build script during dependency-status install; direct repository typecheck also reports pre-existing unrelated errors outside this change.

### DIFF
--- a/src/ui/fonts/useCustomFonts.ts
+++ b/src/ui/fonts/useCustomFonts.ts
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
-import { useEffect } from 'react'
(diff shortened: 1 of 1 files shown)


psiddharthdesign:
@mergifyio queue

PR #407, opened 2026-06-28
fix(cli): report patch scene argument errors

psiddharthdesign:
## Summary\n- return a structured MCP error when patch_scene receives invalid arguments\n- add direct MCP tests for the patch_scene schema and invalid argument path\n\n## Tests\n- pnpm test (in cli/)


psiddharthdesign:
@mergifyio queue

PR #416, opened 2026-06-28
Reject array scene JSON in create command

psiddharthdesign:
## Summary
- reject top-level JSON arrays in the `hypermotion create` command
- add a command test that asserts the error and avoids writing an output file

## Tests
- `pnpm test` from `cli/`


psiddharthdesign:
@mergifyio queue

PR #419, opened 2026-06-28
docs: refresh render scene guidance

psiddharthdesign:
## Summary\n- Update AGENTS.md to show that render_scene can render a provided .hype scene path\n- Refresh the CLI render command comment to match current file-based render support\n\n## Tests\n- pnpm install --frozen-lockfile\n- pnpm test


psiddharthdesign:
@mergifyio queue

PR #420, opened 2026-06-28
Reject array scene JSON in MCP create

psiddharthdesign:
## Summary
- reject top-level array scene JSON in the MCP create_scene handler
- add coverage for JSON-string arrays so MCP validation matches the CLI create command

## Verification
- pnpm --dir cli test

Note: root `pnpm typecheck` currently fails on unrelated app TypeScript errors outside this change.


psiddharthdesign:
@mergifyio queue

PR #424, opened 2026-06-28
test: allow sync env helper callbacks

psiddharthdesign:
## Summary
- widen the CLI withEnvVar helper callback type to allow synchronous callbacks
- add a unit test that verifies sync callbacks still restore the previous env value

## Verification
- pnpm --dir cli test

Note: full root typecheck was checked before this patch and is currently failing on unrelated existing app-level TypeScript errors on origin/main.


psiddharthdesign:
@mergifyio queue

PR #430, opened 2026-06-28
test(cli): share process exit test helper

psiddharthdesign:
## Summary
- add a shared CLI test helper for intercepting process.exit
- use it in open/render command tests so global restoration stays centralized

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #431, opened 2026-06-28
test(cli): cover env restore after async failure

psiddharthdesign:
## Summary\n- add a regression test that verifies withEnvVar restores the original value after an async callback rejects\n\n## Verification\n- pnpm --dir cli test -- --test-name-pattern='withEnvVar'


psiddharthdesign:
@mergifyio queue

PR #434, opened 2026-06-28
test(cli): cover info missing scene error

psiddharthdesign:
## Summary
- add CLI info command coverage for missing scene file errors
- assert the command exits with code 2 and writes the expected stderr prefix

## Tests
- pnpm run build && node --test dist/commands/info.test.js
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #436, opened 2026-06-28
Use SceneSummary type in CLI info paths

psiddharthdesign:
## Summary
- use the exported SceneSummary type in CLI and MCP info handlers
- avoid duplicating the readSceneSummary return type expression at call sites

## Tests
- pnpm --dir cli exec tsc --noEmit
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #437, opened 2026-06-28
Honor encodings in CLI stdout capture helper

psiddharthdesign:
## Summary
- decode captured CLI test output with the provided string chunk encoding
- add a regression test for base64-encoded stdout writes

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #439, opened 2026-06-28
test(cli): cover malformed info_scene files

psiddharthdesign:
## Summary
- Add MCP info_scene coverage for malformed .hype files
- Assert the handler returns a structured MCP error instead of throwing

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #444, opened 2026-06-28
test(cli): reuse process exit helper in create tests

psiddharthdesign:
## Summary
- Reuse the shared process exit helper in the create command test
- Make the helper generic so callers can return captured output

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #446, opened 2026-06-28
test(cli): cover env restoration on sync failure

psiddharthdesign:
## Summary\n- add coverage for withEnvVar restoring the previous value after a synchronous callback throws\n\n## Verification\n- pnpm test


psiddharthdesign:
@mergifyio queue

PR #450, opened 2026-06-28
chore(cli): type stdin chunks in create command

psiddharthdesign:
## Summary
- Type the create command stdin data callback as a string after setting stdin encoding.

## Verification
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #451, opened 2026-06-28
Align render CLI fps limit

psiddharthdesign:
## Summary
- cap the human CLI render fps validation at 120 to match the MCP render tool schema
- add a regression test that rejects 121 fps before locating or launching the desktop app

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #452, opened 2026-06-28
test(cli): allow env helper to unset variables

psiddharthdesign:
## Summary
- Let the CLI test env helper temporarily unset an existing environment variable
- Add coverage that the previous value is restored afterward

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #457, opened 2026-06-28
test(cli): cover missing MCP layer lookup

psiddharthdesign:
## Summary
- Add coverage for the `get_layer` MCP handler when a valid scene does not contain the requested node id.

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #459, opened 2026-06-28
Fix default process exit test helper code

psiddharthdesign:
## Summary\n- make withProcessExitThrow report exitCode 0 when process.exit() is called without an explicit code\n- add a regression test for the default process.exit() case\n\n## Tests\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #460, opened 2026-06-28
Clarify optional camera guidance

psiddharthdesign:
## Summary
- Clarify that create_scene only requires a root frame for renderable scenes
- Keep camera nodes optional unless the design needs camera properties
- Add a regression assertion for the MCP tool description

## Tests
- pnpm --dir cli test -- createScene


psiddharthdesign:
@mergifyio queue

PR #465, opened 2026-06-28
test(cli): cover patch command output

psiddharthdesign:
## Summary\n- add command-level coverage for patching a scene to an alternate output path\n- verify the original scene is unchanged and the patched output is written\n\n## Tests\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #469, opened 2026-06-28
test(cli): cover open command launch path

psiddharthdesign:
## Summary\n- add a focused success-path test for the CLI open command\n- inject the desktop app locator/spawn call so the test can verify launch args without starting Electron\n\n## Tests\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #471, opened 2026-06-28
test(cli): cover stderr encoded writes

psiddharthdesign:
## Summary
- Add stderr coverage for encoded string chunks in the CLI stream capture test helper.

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #473, opened 2026-06-28
test(cli): avoid cast in missing-node validation fixture

psiddharthdesign:
## Summary
- Rework the missing-node validation fixture to mutate the typed sample scene instead of casting through unknown.

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #474, opened 2026-06-28
docs(cli): avoid stale scope version heading

psiddharthdesign:
## Summary
- Rename the CLI README scope heading from a fixed version to a current-scope heading so the quick-reference does not go stale on future releases.

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #478, opened 2026-06-28
Improve CLI env test helper return typing

psiddharthdesign:
## Summary
- Preserve and type the callback return value from the CLI withEnvVar test helper.
- Add sync and async return-value coverage for the helper.

## Tests
- pnpm test (in cli/)


psiddharthdesign:
@mergifyio queue

PR #479, opened 2026-06-28
Tighten scene patch payload types

psiddharthdesign:
## Summary
- restrict setMeta and setNode patch payload types to JSON-safe objects in the CLI scene patch contract

## Tests
- pnpm test (from cli/)


psiddharthdesign:
@mergifyio queue

PR #483, opened 2026-06-28
Tighten process exit test assertion

psiddharthdesign:
## Summary
- replace the process-exit test cast with an assertion helper that narrows the thrown error shape

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #487, opened 2026-06-28
test(cli): assert capability schemas take no args

psiddharthdesign:
## Summary
- make the no-argument capability MCP schemas explicitly declare an empty required list
- add a focused test that keeps both capability schemas aligned with that contract

## Tests
- pnpm test (in cli/)


psiddharthdesign:
@mergifyio queue

PR #490, opened 2026-06-28
fix(cli): handle malformed inspect MCP scenes

psiddharthdesign:
## Summary
- return a structured MCP error when inspect_scene reads malformed .hype data
- add regression coverage for malformed inspect_scene inputs

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #492, opened 2026-06-28
test(cli): cover sync env helper failures

psiddharthdesign:
## Summary
- Add coverage that withEnvVar restores the previous value after a synchronous callback throws.

## Verification
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #494, opened 2026-06-28
test(cli): cover unfiltered track listing

psiddharthdesign:
## Summary
- add MCP query coverage for list_tracks without a node filter
- keep the existing filtered-track assertion against a multi-track fixture

## Tests
- pnpm --dir cli test -- --test-name-pattern "query scene MCP"


psiddharthdesign:
@mergifyio queue

PR #507, opened 2026-06-29
fix(cli): surface plain-text render errors

psiddharthdesign:
## Summary
- preserve plain-text render error sentinels when they are not valid JSON
- add a focused CLI driver test with a fake app writing a text error sentinel

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #512, opened 2026-06-29
fix(cli): handle missing patch scene files

psiddharthdesign:
## Summary\n- return structured MCP errors when patch_scene cannot read the input scene\n- add regression coverage for missing patch scene files\n\n## Tests\n- pnpm --dir cli test -- --test-name-pattern patch_scene


psiddharthdesign:
@mergifyio queue

PR #519, opened 2026-06-29
test(cli): cover malformed query scene inputs

psiddharthdesign:
## Summary
- add malformed .hype coverage for query-scene MCP handlers
- verify handlers return MCP errors consistently for unreadable scene bytes

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #524, opened 2026-06-29
test(cli): cover env unset failure restore

psiddharthdesign:
## Summary
- add coverage that withEnvVar restores an existing variable after a temporarily-unset callback throws
- tidy one extra blank line in the same test file

## Tests
- pnpm test -- --test-name-pattern=withEnvVar


psiddharthdesign:
@mergifyio queue

PR #526, opened 2026-06-29
chore(cli): type MCP layer summaries

psiddharthdesign:
## Summary
- add an explicit LayerSummary type for list_layers MCP responses
- annotate the layer mapper return shape without changing runtime behavior

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #527, opened 2026-06-29
chore(cli): type query scene tool names

psiddharthdesign:
## Summary
- constrain query-scene MCP helper tool-name parameters to the supported query tool names

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #533, opened 2026-06-29
docs: align scene doc storage types

psiddharthdesign:
## Summary
- Update the Y.Doc shape comment in src/scene/doc.ts to match the current Y.Map<unknown>-based storage boundary.
- Remove the stale comment-only any/keyed-map wording from the scene storage diagram.

## Verification
- PASS: pnpm exec eslint src/scene/doc.ts
- PASS: git diff --check
- NOTE: pnpm lint currently fails on unrelated existing React Hooks / refresh lint errors outside this change.
- NOTE: pnpm typecheck currently fails on unrelated existing TypeScript errors outside this change.

### DIFF
--- a/src/scene/doc.ts
+++ b/src/scene/doc.ts
@@ -54,10 +54,10 @@ const DEFAULT_UI_STATE: UiStateSlab = {
  * Y.Doc shape:
  *
  *   doc.getMap('scene')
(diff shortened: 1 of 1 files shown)


psiddharthdesign:
@mergifyio queue

PR #537, opened 2026-06-29
chore(cli): type query scene snapshot

psiddharthdesign:
## Summary
- add a narrow QuerySceneSnapshot type for decoded MCP query scene data
- keep runtime behavior unchanged while documenting the top-level fields the handlers use

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #540, opened 2026-06-29
test(cli): assert render value flags

psiddharthdesign:
## Summary
- extend the headless render driver test to require --format=, --quality=, and --fps= value flags
- keeps coverage aligned with the Electron second-instance argv behavior documented in the driver

## Test
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #541, opened 2026-06-29
Use forced file removal for render cleanup

psiddharthdesign:
## Summary
- replace the render cleanup helper's exists-then-unlink sequence with forced removal
- keep cleanup best-effort while avoiding a small existence-check race window

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #543, opened 2026-06-29
Type inspect command options

psiddharthdesign:
## Summary
- Add an explicit options type for the CLI inspect command action.
- Preserve current JSON-only inspect output behavior.

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #546, opened 2026-06-29
test(cli): narrow invalid track test type

psiddharthdesign:
## Summary
- replace a broad double assertion in the invalid track property test with a local invalid scene type
- keep the negative validation fixture explicit while preserving the intended invalid property id coverage

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #547, opened 2026-06-29
test(cli): cover info default timing

psiddharthdesign:
## Summary\n- add a focused CLI info command test for builder-provided default duration and frame rate output\n\n## Verification\n- pnpm --dir cli exec tsc && node --test cli/dist/commands/info.test.js\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #549, opened 2026-06-29
chore(scene): export scene node alias

psiddharthdesign:
## Summary
- export a SceneNode alias for the scene node union from the public scene types
- keeps imports explicit where Node would collide with the DOM Node type

## Verification
- pnpm -C cli test
- pnpm typecheck (fails on unrelated existing root TS errors; SceneNode export errors are gone)


psiddharthdesign:
@mergifyio queue

PR #551, opened 2026-06-29
test(cli): cover render error fallback

psiddharthdesign:
## Summary
- add a focused CLI driver test for JSON error sentinels that omit a message
- assert the generic fallback error is surfaced and the sentinel is cleaned up

## Verification
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #552, opened 2026-06-29
Harden CLI render error sentinel parsing

psiddharthdesign:
## Summary
- Treat JSON render error sentinel messages as usable only when they are strings
- Add a regression test for non-string JSON sentinel messages falling back cleanly

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #553, opened 2026-06-29
test(cli): cover media default timing fields

psiddharthdesign:
## Summary
- Extend the CLI scene builder media-default test to cover default start time, trim start, loop, duration, and volume fields.

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #555, opened 2026-06-29
Add blank render error sentinel test

psiddharthdesign:
## Summary
- add CLI render driver coverage for blank plain-text error sentinels
- verify blank sentinel files fall back to the generic render failure and are cleaned up

## Tests
- pnpm --dir cli test -- --test-name-pattern "driveHeadlessRender"


psiddharthdesign:
@mergifyio queue

PR #556, opened 2026-06-29
Validate single-camera scene invariant

psiddharthdesign:
## Summary
- Reject authored .hype scenes that contain multiple camera nodes during CLI validation
- Add a focused regression test for the documented single-camera scene constraint

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #561, opened 2026-06-29
chore(cli): type render scene input

psiddharthdesign:
## Summary
- add an explicit inferred type for the render_scene MCP input payload
- route parsed render input through the typed local before applying defaults

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #562, opened 2026-06-29
chore(cli): type open scene input

psiddharthdesign:
## Summary
- add an inferred Zod input type for the open_scene MCP handler
- use the typed parsed input before resolving the scene path

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #563, opened 2026-06-29
Validate malformed scene children fields

psiddharthdesign:
## Summary
- reject malformed scene nodes whose children field is present but not an array
- add CLI validation coverage for the malformed children shape

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #571, opened 2026-06-29
fix(cli): handle malformed MCP patch JSON

psiddharthdesign:
## Summary
- return a structured MCP error when patch_scene receives malformed JSON patch strings
- add a regression test for malformed patch JSON

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #575, opened 2026-06-30
test(cli): cover patch overwrite default

psiddharthdesign:
## Summary
- Add CLI coverage for the default `hypermotion patch` behavior that overwrites the input scene when `--output` is omitted.

## Tests
- `pnpm --dir cli test`


psiddharthdesign:
@mergifyio queue

PR #576, opened 2026-06-30
fix(cli): reject query scene directories

psiddharthdesign:
## Summary
- report directory scene paths as MCP errors before reading query-scene files
- add coverage for directory inputs across list_layers, get_layer, list_tracks, and list_cameras

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #579, opened 2026-06-30
Refactor CLI render error sentinel parsing

psiddharthdesign:
## Summary
- replace the inline JSON sentinel assertion with a narrow type guard
- keep fallback behavior for plain text, blank, and malformed error sentinels unchanged

## Tests
- pnpm --dir cli run build && node --test cli/dist/electron/driver.test.js
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #580, opened 2026-06-30
Handle blank render error messages

psiddharthdesign:
## Summary
- Treat whitespace-only JSON render error sentinel messages as missing details
- Add a CLI driver regression test for blank JSON error messages

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #584, opened 2026-06-30
fix(cli): prune tracks for deleted scene nodes

psiddharthdesign:
## Summary
- remove animation tracks that target nodes deleted by scene patch operations
- add regression coverage that validates the patched scene after deleting a tracked node

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #586, opened 2026-06-30
fix(cli): handle open scene stat failures

psiddharthdesign:
## Summary
- Return a structured MCP error when open_scene cannot stat a scene path after existence checks
- Add focused coverage for the stat-failure path

## Tests
- pnpm --dir cli test -- openScene.test.js


psiddharthdesign:
@mergifyio queue

PR #587, opened 2026-06-30
fix(cli): handle open command stat failures

psiddharthdesign:
## Summary
- catch stat failures in the CLI open command and report a clean user-facing error
- keep the app launch path untouched unless the scene path validates as a file
- add regression coverage for the stat-failure path

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #589, opened 2026-06-30
fix(cli): handle render scene stat failures

psiddharthdesign:
## Summary
- Guard render command scene stat reads so filesystem errors become clean CLI errors
- Add a regression test that verifies render exits before launching when stat fails

## Verification
- pnpm --dir cli test

Note: repository-wide `pnpm typecheck` currently fails on existing app-side TypeScript errors unrelated to this CLI change.


psiddharthdesign:
@mergifyio queue

PR #590, opened 2026-06-30
Handle render_scene stat failures

psiddharthdesign:
## Summary
- Return a structured MCP error when render_scene cannot stat a provided scene path
- Add coverage for the stat-failure path

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #596, opened 2026-06-30
test(cli): reuse MCP text helper in open scene tests

psiddharthdesign:
## Summary
- Reuse the shared MCP text assertion helper in open_scene tests
- Remove duplicated inline content extraction in those assertions

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #598, opened 2026-06-30
fix(cli): reject inspect directory inputs

psiddharthdesign:
## Summary
- reject directory paths before inspect reads scene bytes
- add a command-level regression test for directory inputs

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #602, opened 2026-06-30
Handle MCP render driver failures

psiddharthdesign:
## Summary
- Return MCP error results when render_scene driver execution fails
- Add regression coverage for a failing desktop driver process

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #603, opened 2026-06-30
Tighten CLI stdout helper callback type

psiddharthdesign:
## Summary
- Allow CLI stdout/stderr capture helper callbacks to return any value, matching current `process.stdout.write(...)` test callbacks.
- Keep capture behavior unchanged; the helper still returns only captured output.

## Verification
- `pnpm --dir cli test`
- `pnpm exec eslint cli/src/testUtils/stdout.ts`
- `git diff --check`

## Notes
- `pnpm lint` and `pnpm typecheck` were also run, but both fail on existing unrelated repository issues outside this one-file change.


psiddharthdesign:
@mergifyio queue

PR #611, opened 2026-06-30
Treat whitespace app path overrides as unset

psiddharthdesign:
## Summary
- trim HYPERMOTION_APP_PATH before treating it as an explicit desktop app override
- cover whitespace-only overrides with the existing locator unset behavior

## Tests
- pnpm --dir cli run build && node --test cli/dist/electron/locator.test.js
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #612, opened 2026-06-30
Isolate array scene MCP test output

psiddharthdesign:
## Summary
- Use a unique temp directory for the create_scene array JSON rejection test
- Clean up the temp directory with try/finally

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #616, opened 2026-06-30
test: cover nested stdout capture

psiddharthdesign:
## Summary
- Add regression coverage for nested stdout capture scoping in the CLI test helper
- Verify nested capture output does not leak into the outer capture and restores the original writer

## Tests
- pnpm --dir cli run build && node --test cli/dist/testUtils/stdout.test.js
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #619, opened 2026-06-30
fix(cli): resolve inspect scene paths

psiddharthdesign:
## Summary
- resolve inspect command scene paths before stat/read/error output
- add regression coverage for relative path diagnostics

## Tests
- pnpm --dir cli run build && node --test cli/dist/commands/inspect.test.js
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #620, opened 2026-06-30
Reject fractional render fps

psiddharthdesign:
## Summary
- reject fractional fps values in the CLI render command before launching the app
- add a regression test for fractional fps validation

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #621, opened 2026-06-30
Validate render output parent paths

psiddharthdesign:
## Summary
- fail early when the render output parent path exists but is not a directory
- add CLI coverage for the output-parent file case

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #622, opened 2026-06-30
Cover fractional MCP render fps rejection

psiddharthdesign:
## Summary
- add MCP render_scene coverage for fractional fps rejection
- keep handler/schema behavior aligned with the integer fps contract

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #625, opened 2026-06-30
Tighten MCP test text helper

psiddharthdesign:
## Summary
- add an explicit text-content type guard for the MCP test helper
- keep helper behavior unchanged while narrowing before reading `.text`

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #627, opened 2026-06-30
test(cli): share query scene tool text helper

psiddharthdesign:
## Summary
- add a small query-scene MCP test helper for asserting text tool results
- replace repeated ad hoc text extraction in that test file

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #628, opened 2026-06-30
test(cli): cover non-string scene child references

psiddharthdesign:
## Summary
- add a focused validation regression test for non-string scene child references in saved .hype data

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #629, opened 2026-06-30
test(cli): cover MCP patch in-place overwrite

psiddharthdesign:
## Summary
- add MCP patch_scene coverage for default in-place overwrite behavior
- verify the patched scene replaces the original when output is omitted and live apply is disabled

## Tests
- pnpm --dir cli test -- patchScene.test.js
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #630, opened 2026-06-30
Handle validate stat failures cleanly

psiddharthdesign:
## Summary\n- handle CLI validate stat failures through the existing read-error path\n- add regression coverage for stat failures before scene reads\n\n## Verification\n- pnpm --dir cli run build && node --test cli/dist/commands/validate.test.js\n- pnpm test\n- pnpm exec eslint cli/src/commands/validate.ts cli/src/commands/validate.test.ts\n\nNote: root pnpm lint still fails on pre-existing unrelated app-wide issues (unused vars / React hook rules outside this patch).

### DIFF
--- a/cli/src/commands/validate.test.ts
+++ b/cli/src/commands/validate.test.ts
@@ -361,6 +361,41 @@ test('validate command reports directories before reading scene bytes', async ()
   }
 })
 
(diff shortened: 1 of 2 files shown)


psiddharthdesign:
@mergifyio queue

PR #631, opened 2026-06-30
Handle patch command directory inputs

psiddharthdesign:
## Summary
- Resolve patch command scene paths before reading
- Report directory scene inputs as non-files before patch parsing
- Add CLI regression coverage for patch directory inputs

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #632, opened 2026-06-30
Handle render_scene output stat failures

psiddharthdesign:
## Summary
- convert output directory stat failures in render_scene into MCP error responses
- add regression coverage for the output directory stat failure path

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #633, opened 2026-06-30
Cover validate read failures

psiddharthdesign:
## Summary
- add CLI validate command coverage for readFileSync failures after stat succeeds
- assert the command reports the read error through the existing read-failure path

## Tests
- pnpm --dir cli run build && node --test cli/dist/commands/validate.test.js
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #635, opened 2026-06-30
Handle inspect_scene directory inputs

psiddharthdesign:
## Summary
- return a clear MCP error when inspect_scene receives a directory path
- add coverage for directory inputs in inspect_scene

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #639, opened 2026-06-30
docs(cli): refresh create command layout example

psiddharthdesign:
## Summary
- refresh the CLI create command's inline scene example to use the current auto-layout guidance

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #642, opened 2026-06-30
Handle info command stat failures

psiddharthdesign:
## Summary\n- route info command stat failures through the existing read-error message and exit code\n- add coverage for the stat failure path\n\n## Tests\n- pnpm --dir cli run build && node --test cli/dist/commands/info.test.js\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #648, opened 2026-06-30
Respect documented verbose flag value

psiddharthdesign:
## Summary
- Treat HYPERMOTION_VERBOSE as enabled only when set to `1`, matching the CLI docs
- Add a driver regression test for `HYPERMOTION_VERBOSE=0`

## Tests
- `pnpm --dir cli run build && node --test cli/dist/electron/driver.test.js`
- `pnpm --dir cli test`


psiddharthdesign:
@mergifyio queue

PR #650, opened 2026-06-30
fix(cli): report output directory stat failures

psiddharthdesign:
## Summary
- catch render output directory stat failures and report a clean CLI error
- add regression coverage that the desktop app is not launched in that case

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #652, opened 2026-06-30
Normalize render CLI option values

psiddharthdesign:
## Summary
- Trim explicit render format and quality values before validation
- Trim fps input before numeric parsing
- Add a regression test for padded render option values

## Tests
- pnpm test

Note: whole-repo `pnpm typecheck` was checked before the change and currently fails on unrelated app-side type errors on `main`.


psiddharthdesign:
@mergifyio queue

PR #656, opened 2026-06-30
test(cli): cover null scene JSON in create command

psiddharthdesign:
## Summary
- Add CLI create command regression coverage for top-level JSON null input
- Verify the command rejects null like other non-object scene JSON and does not write an output file

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #661, opened 2026-07-01
test(cli): escape query scene path regexes

psiddharthdesign:
## Summary
- Escape temporary scene paths before embedding them in query-scene regex assertions
- Avoid regex metacharacters in temp paths causing false test failures

## Tests
- pnpm test (in cli/)


psiddharthdesign:
@mergifyio queue

PR #667, opened 2026-07-01
Trim render scene path input

psiddharthdesign:
## Summary
- trim padded CLI --scene values before resolving and validating render scene paths
- add command coverage for padded scene path input

## Tests
- pnpm --dir cli test -- --test-name-pattern "render command" (runs full CLI suite; 279 passed)

## Notes
- pnpm typecheck currently fails on unrelated app-wide TypeScript errors outside this CLI change.
- pnpm lint currently fails on unrelated app-wide lint errors outside this CLI change.

### DIFF
--- a/cli/src/commands/render.test.ts
+++ b/cli/src/commands/render.test.ts
@@ -147,6 +147,27 @@ test('render command trims padded format, quality, and fps values', async () => 
   }
 })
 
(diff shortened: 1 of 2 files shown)


psiddharthdesign:
@mergifyio queue

PR #669, opened 2026-07-01
fix(cli): clarify query scene malformed file errors

psiddharthdesign:
## Summary\n- report malformed .hype contents from query-scene MCP tools as invalid scene files instead of read failures\n- update the existing malformed-file query scene test expectation\n\n## Tests\n- pnpm --dir cli run build && node --test cli/dist/mcp/queryScene.test.js\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #671, opened 2026-07-01
chore(cli): tighten stdout capture callback type

psiddharthdesign:
## Summary
- tighten the CLI stdout/stderr capture helper callback type to `void`
- keep the helper API aligned with its side-effect-only usage in tests

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #675, opened 2026-07-01
Validate scene parent field types

psiddharthdesign:
## Summary
- reject malformed saved scene nodes whose parent field is neither a string nor null
- add validator coverage for non-string parent fields

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #677, opened 2026-07-01
Resolve CLI create output paths consistently

psiddharthdesign:
## Summary
- resolve the CLI create output path once and use that path for mkdir, write, errors, and success output
- tighten create-command tests to assert the exact written path

## Verification
- pnpm --dir cli test
- pnpm test

Note: pnpm lint was attempted but the current app lint baseline has pre-existing errors outside this change (for example React hook lint errors in src/render/RenderWindowApp.tsx, src/render3d/ThreeSceneViewport.tsx, src/ui/Canvas.tsx, and others).


psiddharthdesign:
@mergifyio queue

PR #679, opened 2026-07-01
test(cli): cover all render option guards

psiddharthdesign:
## Summary
- Extend the CLI render option guard test to assert every declared supported format and quality is accepted.

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #683, opened 2026-07-01
fix(cli): trim MCP render scene paths

psiddharthdesign:
## Summary
- Trim optional MCP render_scene scene paths before resolving them
- Add coverage that padded scene paths are forwarded cleanly to the render driver

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #685, opened 2026-07-01
chore(cli): use typed scene guard sets

psiddharthdesign:
## Summary
- add readonly membership sets for scene node kinds and keyframe property ids
- use those sets in validation type guards instead of widening typed arrays

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #688, opened 2026-07-01
fix(cli): report render output mkdir failures

psiddharthdesign:
## Summary
- return a structured MCP error when render_scene cannot create the output directory
- add regression coverage for output directory creation failures

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #689, opened 2026-07-01
chore(cli): split validate stat check

psiddharthdesign:
## Summary
- split validate command stat/non-file handling from scene byte reads
- keep directory inputs on the explicit non-file error path

## Verification
- pnpm --dir cli test

Note: repo-level `pnpm typecheck` currently fails on unrelated existing app TypeScript errors outside this CLI change.


psiddharthdesign:
@mergifyio queue

PR #691, opened 2026-07-01
test(cli): cover extensionless render fallback

psiddharthdesign:
## Summary
- add a render option assertion for output paths that end with a trailing dot
- keeps extensionless or unknown output paths pinned to the mp4 fallback behavior

## Tests
- pnpm run build && node --test dist/renderOptions.test.js
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #693, opened 2026-07-01
test(cli): cover trimmed open scene paths

psiddharthdesign:
## Summary\n- trim padded scene arguments in the CLI open command before resolving the path\n- add a regression test that verifies the normalized path is passed to the desktop app\n\n## Tests\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #695, opened 2026-07-01
fix(cli): trim MCP render output paths

psiddharthdesign:
## Summary
- Trim padded MCP render output paths before resolving them
- Add a regression test that verifies the desktop driver receives the cleaned output path

## Tests
- pnpm --dir cli test -- --test-name-pattern render_scene


psiddharthdesign:
@mergifyio queue

PR #696, opened 2026-07-01
Normalize invalid fps error output

psiddharthdesign:
## Summary
- print the normalized fps input when render command validation fails
- add a regression test for padded invalid fps values

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #699, opened 2026-07-01
fix(cli): trim create scene output paths

psiddharthdesign:
## Summary
- trim whitespace around create_scene MCP output paths before validation and writing
- add a regression test covering padded output paths

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #707, opened 2026-07-01
Isolate create scene JSON rejection tests

psiddharthdesign:
## Summary
- Use per-test temporary directories for null and malformed create_scene JSON rejection tests
- Clean those temp directories after assertions to avoid shared tmp filename collisions

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #708, opened 2026-07-01
test(cli): cover dotted render output names

psiddharthdesign:
## Summary
- add coverage for render format inference when output filenames contain extra dots before the extension

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #710, opened 2026-07-01
Validate keyframe ids and times

psiddharthdesign:
## Summary
- validate keyframe ids and times in saved scene validation
- add focused validator coverage for malformed keyframe metadata

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #714, opened 2026-07-01
Validate create_scene output before building

psiddharthdesign:
## Summary
- Validate the MCP create_scene output path before building scene bytes
- Add a regression test for invalid output paths taking precedence over scene build errors

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #715, opened 2026-07-01
Reject empty render output paths

psiddharthdesign:
## Summary
- reject whitespace-only render output paths in the CLI before resolving them
- return a clear MCP error for whitespace-only render_scene output values
- cover both paths with focused tests

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #716, opened 2026-07-01
Harden render option guards

psiddharthdesign:
## Summary
- Accept unknown inputs in render format and quality guards before narrowing to supported values
- Cover non-string guard inputs in the existing render options test

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #719, opened 2026-07-01
test(cli): cover combined render output suffixes

psiddharthdesign:
## Summary\n- Add render format inference coverage for output paths with both query and fragment suffixes.\n\n## Verification\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #723, opened 2026-07-01
test(cli): accept soft text animation smoothing

psiddharthdesign:
## Summary
- Allow CLI-authored text animation configs to use the app-supported `soft` smoothing value
- Add a scene builder regression test that preserves `soft` smoothing on text animation tracks

## Tests
- `pnpm --dir cli test`


psiddharthdesign:
@mergifyio queue

PR #724, opened 2026-07-01
test(cli): cover stale output cleanup on render errors

psiddharthdesign:
## Summary
- extend the JSON error sentinel driver test to seed stale output before render
- assert stale output is removed when the render reports an error

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #725, opened 2026-07-01
Update CLI render flow docs

psiddharthdesign:
## Summary
- Update the packaged CLI README to describe the current hidden render-window export flow instead of the older off-screen/current-scene wording.

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #727, opened 2026-07-01
fix(cli): trim validate scene paths

psiddharthdesign:
## Summary
- Trim padded scene paths in the validate_scene MCP handler before filesystem reads
- Add coverage for padded validate_scene paths

## Verification
- pnpm --dir cli test

Note: pnpm typecheck was attempted before this CLI-only change and currently fails on unrelated existing app-wide TypeScript errors.


psiddharthdesign:
@mergifyio queue

PR #732, opened 2026-07-01
docs(cli): clarify saved scene render behavior

psiddharthdesign:
## Summary
- Clarify that the CLI renders the current desktop scene only when `--scene` is omitted.

## Tests
- pnpm test

Note: `pnpm typecheck` was checked before the docs edit and still has unrelated baseline errors on `origin/main`.


psiddharthdesign:
@mergifyio queue

PR #734, opened 2026-07-01
test(cli): keep render option guard coverage in sync

psiddharthdesign:
## Summary
- use exported render format and quality lists in the render option guard test
- keeps the supported-value assertions in sync with the CLI option source of truth

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #738, opened 2026-07-01
fix(cli): trim create output paths

psiddharthdesign:
## Summary
- Trim padded output paths in the CLI create command before resolving the destination
- Add a regression test covering padded create output arguments

## Verification
- pnpm test
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #741, opened 2026-07-01
fix(cli): reject malformed active camera ids

psiddharthdesign:
## Summary
- reject non-null, non-string activeCameraId values during scene validation
- add a regression test for malformed saved scene data

## Tests
- pnpm --dir cli test
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #742, opened 2026-07-01
chore(cli): readonly query schema type

psiddharthdesign:
## Summary\n- Mark the shared query-scene MCP string schema helper as readonly.\n\n## Verification\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #743, opened 2026-07-01
Validate scene root id type

psiddharthdesign:
## Summary
- report non-string scene.root values as an explicit validation error
- add CLI scene validation coverage for malformed root ids

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #745, opened 2026-07-01
Validate scene node entry objects

psiddharthdesign:
## Summary
- report a direct validation error when a scene node map entry is not an object
- add regression coverage for malformed node entries

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #746, opened 2026-07-01
test(cli): cover padded render output suffixes

psiddharthdesign:
## Summary\n- add coverage for render format inference when output paths include both padding and URL-style suffixes\n\n## Tests\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #747, opened 2026-07-01
Reject non-object scene sections

psiddharthdesign:
## Summary
- report an explicit validation error when a scene section entry is not an object
- update the existing section validation test expectation

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #748, opened 2026-07-01
fix(cli): validate section label fields

psiddharthdesign:
## Summary
- validate section name and color fields in saved .hype scene sections
- add regression coverage for malformed section labels

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #749, opened 2026-07-01
fix(cli): reject non-object track entries

psiddharthdesign:
## Summary
- add a direct validateScene error when a track map entry is not an object
- cover malformed saved track entries with a Yjs document regression test

## Tests
- pnpm --dir cli test
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #751, opened 2026-07-01
fix(cli): validate scene collection shapes

psiddharthdesign:
## Summary
- report explicit validation errors when top-level scene collections are non-object values
- add a focused validator regression test for malformed nodes/tracks/sections containers

## Tests
- pnpm test -- --test-name-pattern "validateScene rejects non-object scene collections|validateScene rejects non-string root ids"


psiddharthdesign:
@mergifyio queue

PR #754, opened 2026-07-01
fix(cli): clarify non-string child validation

psiddharthdesign:
## Summary
- report non-string scene child references with a specific validation error
- cover the CLI validate command and lower-level validator expectation

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #758, opened 2026-07-02
Tighten render option guards

psiddharthdesign:
## Summary
- replace render option guard array assertions with readonly lookup sets
- keep supported format and quality exports unchanged

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #760, opened 2026-07-02
test(cli): trim query scene paths

psiddharthdesign:
## Summary
- Trim padded scene paths before MCP query scene tools read .hype files
- Add coverage for padded query scene paths

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #761, opened 2026-07-02
Document relative render output paths

psiddharthdesign:
## Summary
- Clarify that the MCP render_scene output path accepts absolute or relative paths
- Add a schema test so the documented output path contract stays in sync

## Tests
- pnpm --dir cli test
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #764, opened 2026-07-02
test: cover render format URL suffix fallback

psiddharthdesign:
## Summary
- Add CLI render format inference coverage for query/hash suffixes that look like supported extensions.

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #765, opened 2026-07-02
Trim scene names in info output

psiddharthdesign:
## Summary
- trim padded scene names in human-readable `hypermotion info` output
- add CLI coverage for padded scene names

## Tests
- `pnpm --dir cli test`


psiddharthdesign:
@mergifyio queue

PR #771, opened 2026-07-02
fix(cli): trim info scene paths

psiddharthdesign:
## Summary
- Trim surrounding whitespace from `hypermotion info` scene path arguments before resolving them
- Report whitespace-only scene paths as a clear `[info] scene path is required` error
- Add focused CLI coverage for padded and empty info scene paths

## Tests
- `pnpm --dir cli run build && pnpm --dir cli exec node --test dist/commands/info.test.js`
- `pnpm test`

### DIFF
--- a/cli/src/commands/info.test.ts
+++ b/cli/src/commands/info.test.ts
@@ -292,6 +292,57 @@ test('info command trims padded scene names', async () => {
   }
 })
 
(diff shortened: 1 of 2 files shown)


psiddharthdesign:
@mergifyio queue

PR #773, opened 2026-07-02
Cover unknown extension-only render outputs

psiddharthdesign:
## Summary
- Add a CLI render option test covering unknown extension-only output names such as .mov.
- Locks in the existing fallback behavior to MP4 for unsupported extensions.

## Tests
- pnpm --dir cli run build && node --test cli/dist/renderOptions.test.js
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #776, opened 2026-07-02
Cover malformed render output fallbacks

psiddharthdesign:
## Summary
- add render option tests for whitespace-only and suffix-only output paths falling back to mp4

## Tests
- pnpm --dir cli test
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #782, opened 2026-07-02
fix(cli): reject primitive patch json

psiddharthdesign:
## Summary
- reject primitive or null patch JSON before applying CLI patch operations
- add coverage that the input scene is left unchanged for primitive patch JSON

## Tests
- pnpm --dir cli test -- --test-name-pattern='patch command'
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #784, opened 2026-07-02
chore(cli): type MCP text fallback results

psiddharthdesign:
## Summary
- add a typed helper for MCP text fallback/error responses
- remove repeated inline literal response construction in the MCP server fallback path

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #786, opened 2026-07-02
Clarify CLI image fill fit type

psiddharthdesign:
## Summary
- Add an explicit CLI type alias for image fill fit values
- Keep image fill fit values distinct from media node fit values

## Verification
- pnpm test
- pnpm --dir cli test

Note: root `pnpm typecheck` still fails on existing app-wide TypeScript errors unrelated to this CLI type-only change.


psiddharthdesign:
@mergifyio queue

PR #803, opened 2026-07-02
fix(cli): trim inspect scene paths

psiddharthdesign:
## Summary
- Trim padded scene path arguments in the inspect command before resolving them
- Reject blank inspect scene paths with the same explicit error style as related commands
- Add focused inspect command coverage for padded and blank scene path inputs

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #806, opened 2026-07-02
Handle blank inspect scene paths

psiddharthdesign:
## Summary
- Trim and resolve inspect_scene MCP paths before reading
- Return a clear MCP error for blank inspect_scene paths
- Add coverage for the blank path case

## Tests
- pnpm --dir cli test -- inspectScene
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #810, opened 2026-07-02
fix(cli): trim patch scene paths

psiddharthdesign:
## Summary
- Trim patch command scene paths before resolving them
- Reject blank patch scene paths with a clear CLI error
- Add focused tests for padded and blank patch scene inputs

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #812, opened 2026-07-02
test: cover mixed render path suffixes

psiddharthdesign:
## Summary\n- Add CLI render format inference coverage for paths containing both fragment and query-like suffix text\n- Confirm suffix text after a fragment does not make a hidden extension look renderable\n\n## Tests\n- pnpm --dir cli run build && node --test cli/dist/renderOptions.test.js\n- pnpm test


psiddharthdesign:
@mergifyio queue

PR #817, opened 2026-07-02
Export patch operation type

psiddharthdesign:
## Summary
- Export a PatchOperationType union from the existing PATCH_OPERATION_TYPES constant for CLI scene patch consumers.

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #819, opened 2026-07-02
test: cover query-like render output names

psiddharthdesign:
## Summary
- Add coverage for output paths where a query-like suffix contains a supported extension after `?`.
- Confirms render format inference keeps falling back to `mp4` for suffix-only extensions.

## Tests
- `pnpm --dir cli run build && node --test cli/dist/renderOptions.test.js`
- `pnpm test`


psiddharthdesign:
@mergifyio queue

PR #820, opened 2026-07-02
test(cli): cover inspect read failures

psiddharthdesign:
## Summary
- add CLI inspect command coverage for the path where stat succeeds but file reading fails

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #823, opened 2026-07-02
test(cli): clarify inspected scene object type

psiddharthdesign:
## Summary
- Rename the broad inspected scene test alias to PlainSceneObject
- Keep scene map and nested keyframe assertions aligned with that alias

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #828, opened 2026-07-02
test(cli): cover parent directory format inference

psiddharthdesign:
## Summary
- add a regression test that render format inference ignores supported extensions in parent directory names

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #836, opened 2026-07-02
test(cli): cover non-string render option guards

psiddharthdesign:
## Summary\n- Add explicit CLI render option guard coverage for null and object inputs.\n\n## Verification\n- pnpm --dir cli test\n\n## Notes\n- Repo-wide pnpm typecheck was also run, but it currently fails in unrelated app/editor files (for example src/anim/engine.ts, src/export/encodeGif.ts, and UI components).


psiddharthdesign:
@mergifyio queue

psiddharthdesign:
@mergifyio queue

PR #840, opened 2026-07-03
test(cli): cover encoded stream write callbacks

psiddharthdesign:
## Summary
- add CLI stdout/stderr capture tests for write callbacks passed after explicit encodings

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #843, opened 2026-07-03
Trim create command source paths

psiddharthdesign:
## Summary
- Trim padded --from values in the create command before reading JSON scene files
- Add a regression test for padded JSON source paths

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #851, opened 2026-07-03
chore(cli): mark render scene paths non-empty

psiddharthdesign:
## Summary
- Add minLength metadata for render_scene output and scene string inputs
- Extend the MCP schema test to assert those non-empty path hints

## Tests
- pnpm --dir cli run build && node --test cli/dist/mcp/renderScene.test.js
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #862, opened 2026-07-03
Trim patch command output paths

psiddharthdesign:
## Summary
- trim padded `patch --output` values before resolving them
- add CLI regression coverage for padded patch output paths

## Verification
- `pnpm --dir cli test -- cli/src/commands/patch.test.ts` (365 tests passed)
- `pnpm typecheck` (fails on existing app-wide TypeScript errors unrelated to this CLI change)


psiddharthdesign:
@mergifyio queue

PR #871, opened 2026-07-03
test(cli): cover hidden output suffix inference

psiddharthdesign:
## Summary
- add CLI render format inference coverage for hidden output filenames with supported suffixes

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #874, opened 2026-07-04
chore(cli): declare query scene path schema length

psiddharthdesign:
## Summary
- declare `minLength: 1` on query scene MCP `scene` path schema metadata
- assert the schema bound in the query scene MCP tests

## Verification
- `PATH="/Users/siddharthponnapalli/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH" pnpm test`


psiddharthdesign:
@mergifyio queue

PR #878, opened 2026-07-06
Clarify installer signing comment

psiddharthdesign:
## Summary
- Clarify that the installer downloads an unsigned/not-notarized DMG before locally re-applying an ad-hoc signature.

## Verification
- `bash -n install.sh`


psiddharthdesign:
@Mergifyio queue

PR #879, opened 2026-07-07
Reject blank render scene paths

psiddharthdesign:
## Summary
- Reject an explicitly blank `--scene` value in `hypermotion render` instead of silently rendering the current desktop scene.
- Add CLI coverage for the validation path before app launch.

## Verification
- `PATH="/Users/siddharthponnapalli/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH" pnpm --dir cli test` (passes, 369 tests)

## Notes
- Broader `pnpm typecheck` and `pnpm lint` currently fail on pre-existing app-wide TypeScript/React lint issues outside this CLI change.

### DIFF
--- a/cli/src/commands/render.test.ts
+++ b/cli/src/commands/render.test.ts
@@ -179,6 +179,23 @@ test('render command trims padded scene paths', async () => {
   }
 })
 
+test('render command rejects empty explicit scene paths before launching the app', async () => {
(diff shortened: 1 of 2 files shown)


psiddharthdesign:
@mergifyio queue

PR #880, opened 2026-07-07
chore(cli): declare inspect scene path schema length

psiddharthdesign:
## Summary
- add minLength metadata to the inspect_scene MCP scene path schema
- update the schema assertion to keep the MCP tool contract covered

## Tests
- pnpm --dir cli test -- --test-name-pattern=inspect_scene


psiddharthdesign:
@mergifyio queue

PR #882, opened 2026-07-07
Tighten patch JSON parse typing

psiddharthdesign:
## Summary
- parse patch command JSON as unknown before the existing top-level validation
- keep the existing patch application behavior unchanged after validation

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #884, opened 2026-07-07
chore(cli): declare open scene path schema length

psiddharthdesign:
## Summary
- declare a non-empty string length in the open_scene MCP input schema
- update the existing schema assertion to match

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #898, opened 2026-07-07
fix(cli): reject blank create source paths

psiddharthdesign:
## Summary\n- reject whitespace-only `hypermotion create --from` values before reading stdin\n- add regression coverage that no scene file is written for blank source paths\n\n## Tests\n- `pnpm --dir cli test`


psiddharthdesign:
@mergifyio queue

PR #902, opened 2026-07-07
chore(cli): mark query scene paths nonblank

psiddharthdesign:
## Summary
- advertise the nonblank scene-path pattern on query-scene MCP tool schemas
- cover the schema metadata in the existing query-scene MCP test

## Verification
- pnpm --dir cli test -- --test-name-pattern="query scene MCP tool schemas" (runs CLI build and test suite; 380 passed)

Note: root pnpm typecheck currently fails on existing app TypeScript errors outside this CLI-only change.


psiddharthdesign:
@mergifyio queue

PR #906, opened 2026-07-07
Handle Windows separators in render format inference

psiddharthdesign:
## Summary
- normalize backslashes before inferring render output extensions
- add coverage for Windows-style parent directories and hidden extension-only filenames

## Tests
- pnpm --dir cli test
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #908, opened 2026-07-07
fix(cli): wait for error sentinel contents

psiddharthdesign:
## Summary
- avoid consuming zero-byte render error sentinel files before the renderer finishes writing them
- keep existing JSON/text error sentinel fallback behavior intact

## Verification
- pnpm --dir cli test

Note: root `pnpm typecheck` currently fails on unrelated main branch errors outside this change.


psiddharthdesign:
@mergifyio queue

PR #909, opened 2026-07-07
test(cli): cover capture restore after failed writes

psiddharthdesign:
## Summary\n- add coverage for stdout/stderr capture helpers restoring the original writer after a callback writes and then throws\n\n## Verification\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #910, opened 2026-07-07
chore(cli): tighten render option guards

psiddharthdesign:
## Summary\n- type render option lookup sets as strings\n- remove guard-time assertions for untrusted render format and quality inputs\n\n## Verification\n- pnpm --dir cli test\n\nNote: root pnpm typecheck currently fails on existing app-wide TypeScript errors outside this CLI helper.


psiddharthdesign:
@mergifyio queue

PR #915, opened 2026-07-07
Clarify render FPS limit test name

psiddharthdesign:
## Summary
- Clarify the render command FPS limit test name so it refers to the render limit instead of the MCP limit.

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #916, opened 2026-07-07
Tighten render command mkdir test double

psiddharthdesign:
## Summary
- Replace the render command mkdir failure test cast with a typed `fs.mkdirSync` test double.
- Preserve the test behavior while keeping the mock aligned with the injected dependency type.

## Verification
- `pnpm --dir cli test`


psiddharthdesign:
@mergifyio queue

PR #919, opened 2026-07-07
Handle URL markers in render output basename parsing

psiddharthdesign:
## Summary
- Strip URL-style query/hash markers from the output basename instead of the full path when inferring render format.
- Add regression coverage for parent directories containing `?` or `#`.

## Verification
- `pnpm --dir cli test -- --test-name-pattern='inferRenderFormatFromPath|render option guards'` (builds CLI and passed full CLI suite: 384 tests)
- `pnpm typecheck` (fails on existing app-side TypeScript errors unrelated to this change)


psiddharthdesign:
@mergifyio queue

PR #923, opened 2026-07-07
Export render scene dependency type

psiddharthdesign:
## Summary
- Export the MCP render_scene dependency type for direct type-only use in tests
- Replace parameter inference in the render_scene test with the explicit exported type

## Tests
- pnpm --dir cli test -- renderScene.test.js


psiddharthdesign:
@mergifyio queue

PR #925, opened 2026-07-07
Tighten scene validation malformed data tests

psiddharthdesign:
## Summary
- Exercise malformed persisted scene data by mutating the Yjs document directly
- Remove invalid typed scene aliases and broad casts from validation tests

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #926, opened 2026-07-07
Tidy query scene schema test typing

psiddharthdesign:
## Summary\n- add a local schema-property type alias in the query scene MCP schema test\n- replace repeated ad hoc Record casts with the alias\n\n## Tests\n- pnpm --dir cli run build && node --test --test-concurrency=1 cli/dist/mcp/queryScene.test.js\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #927, opened 2026-07-07
Tighten validate command malformed scene test typing

psiddharthdesign:
## Summary
- add a local persisted scene input alias/helper for malformed validate command fixtures
- replace repeated inline double-casts in three validation tests

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #930, opened 2026-07-07
Tighten render scene schema test typing

psiddharthdesign:
## Summary
- Mark the render_scene schema test helper fields as readonly
- Keep enum typing readonly to match schema constants used by the assertions

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #932, opened 2026-07-07
Type serve command options

psiddharthdesign:
## Summary
- Replace the inline serve command action options shape with a named ServeCommandOptions type for consistency with other CLI commands.

## Tests
- pnpm --dir cli run build && node --test cli/dist/commands/serve.test.js
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #941, opened 2026-07-07
chore(cli): align info scene path schema

psiddharthdesign:
## Summary
- Add the non-whitespace path pattern to the info_scene MCP schema
- Update the schema assertion test to match the runtime's whitespace-only path rejection

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #942, opened 2026-07-07
chore(cli): type render success sentinel

psiddharthdesign:
## Summary
- extract the render success sentinel JSON shape into a typed guard
- keep existing sentinel validation behavior unchanged

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #943, opened 2026-07-07
Test render format inference with URL suffixes

psiddharthdesign:
## Summary
- add CLI render command coverage for output paths with URL-style query/hash suffixes
- verify the command still infers the actual file extension before invoking the driver

## Verification
- pnpm --dir cli run build && node --test --test-concurrency=1 cli/dist/commands/render.test.js cli/dist/renderOptions.test.js
- pnpm test

Note: pnpm typecheck was run and fails on existing app-level TypeScript errors unrelated to this test-only change.

### DIFF
--- a/cli/src/commands/render.test.ts
+++ b/cli/src/commands/render.test.ts
@@ -77,6 +77,25 @@ test('render command infers formats case-insensitively from output extensions', 
   }
 })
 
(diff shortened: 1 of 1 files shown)


psiddharthdesign:
@mergifyio queue

PR #952, opened 2026-07-08
chore(cli): tighten render option set typing

psiddharthdesign:
## Summary
- tighten the internal render option sets to derive their TypeScript shape directly from Set<string> construction
- keep render format/quality guard behavior unchanged

## Verification
- pnpm --dir cli test (386 passed)
- pnpm typecheck currently fails on pre-existing app-level TypeScript errors outside this CLI change


psiddharthdesign:
@mergifyio queue

PR #955, opened 2026-07-08
docs(cli): mention uiState in scene builder layout

psiddharthdesign:
## Summary
- Update the CLI scene builder layout comment to include the seeded `uiState` map.
- Keep the comment aligned with the existing editor compatibility behavior.

## Tests
- `pnpm --dir cli test`


psiddharthdesign:
@mergifyio queue

PR #957, opened 2026-07-08
docs(cli): include active camera in create example

psiddharthdesign:
## Summary
- Add the missing activeCameraId field to the CLI create command's scene JSON example so it matches the current camera authoring guidance.

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #959, opened 2026-07-08
chore(cli): filter persisted child ids by type

psiddharthdesign:
## Summary
- replace two `Y.Array` child-id casts with a shared string-filtering helper
- keep CLI scene patch traversal typed against persisted data instead of trusting mixed arrays

## Testing
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #961, opened 2026-07-08
test(cli): cover empty open scene schema path

psiddharthdesign:
## Summary
- add MCP coverage for open_scene rejecting an empty schema scene path

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #962, opened 2026-07-08
test(cli): use shared MCP text assertion

psiddharthdesign:
## Summary
- use the existing assertToolText helper in patch_scene MCP tests
- make patch_scene MCP test assertions fail clearly when text content is missing or malformed

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #982, opened 2026-07-08
Reject blank patch output paths

psiddharthdesign:
## Summary
- Reject whitespace-only `hypermotion patch --output` values with a clear usage error
- Add a regression test for blank patch output paths

## Verification
- `pnpm test`


psiddharthdesign:
@mergifyio queue

PR #985, opened 2026-07-08
fix(cli): reject blank patch input paths

psiddharthdesign:
## Summary
- Reject whitespace-only patch JSON paths in the CLI patch command
- Trim the patch path before reading it, matching scene/output path handling
- Add coverage for the blank --from path case

## Tests
- pnpm --dir cli run build && node --test --test-concurrency=1 cli/dist/commands/patch.test.js
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #986, opened 2026-07-08
test(cli): cover equals-form scene render option

psiddharthdesign:
## Summary
- add CLI regression coverage for passing saved scene paths as `--scene=<path>`
- keep the change test-only and scoped to render command parsing

## Verification
- `pnpm --dir cli test`


psiddharthdesign:
@mergifyio queue

PR #987, opened 2026-07-08
Normalize query layer child ids

psiddharthdesign:
## Summary
- normalize `list_layers` child ids to strings in the CLI MCP query response
- add coverage for malformed child entries being filtered from the public payload

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #988, opened 2026-07-08
Handle missing parent validation safely

psiddharthdesign:
## Summary
- stop scene validation parent-cycle traversal when it reaches an already-missing parent
- add a regression test for malformed scenes with missing parent ids

## Tests
- pnpm --dir cli run build && node --test --test-concurrency=1 cli/dist/scene/build.test.js
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #989, opened 2026-07-08
fix(cli): validate query scene paths in schema

psiddharthdesign:
## Summary
- align query scene MCP path parsing with its advertised non-empty schema
- update blank scene path assertions to expect schema validation errors

## Tests
- pnpm --dir cli run build && node --test --test-concurrency=1 cli/dist/mcp/queryScene.test.js
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #990, opened 2026-07-08
Tighten render scene mkdir dependency type

psiddharthdesign:
## Summary
- Align the render_scene MCP dependency type for mkdirSync with the MakeDirectoryOptions shape used by the handler and test doubles.

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1002, opened 2026-07-08
Tighten env test cleanup

psiddharthdesign:
## Summary
- wrap the `withEnvVar` return-value tests in cleanup guards
- ensure temporary test env vars are removed even if assertions fail

## Tests
- `pnpm --dir cli test`


psiddharthdesign:
@mergifyio queue

PR #1003, opened 2026-07-08
chore(cli): narrow render command test deps

psiddharthdesign:
## Summary
- narrow render command dependency hook types for injected fs helpers
- remove now-unnecessary casts in render command tests

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1006, opened 2026-07-08
Handle directory-like render output paths

psiddharthdesign:
## Summary
- default render format inference to mp4 when the output path ends with a path separator
- add coverage for Unix and Windows-style directory-like output paths

## Tests
- pnpm --dir cli run build && node --test cli/dist/renderOptions.test.js
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #1008, opened 2026-07-08
chore(cli): reject directory create outputs

psiddharthdesign:
## Summary
- reject existing directories as invalid hypermotion create output paths before writing
- add a regression test for directory output paths

## Verification
- pnpm --dir cli test -- create.test.js
- pnpm test

Note: pnpm typecheck still fails on existing app-wide TypeScript errors outside this CLI change.


psiddharthdesign:
@mergifyio queue

PR #1010, opened 2026-07-08
test(cli): cover render scene stat race

psiddharthdesign:
## Summary
- add CLI render command coverage for a saved scene path that disappears between existence and stat checks
- verify the command reports the read failure before launching the desktop app

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1014, opened 2026-07-08
test(cli): cover render format validation order

psiddharthdesign:
## Summary
- add a focused CLI render test that unsupported explicit formats fail before output directory creation or app lookup

## Verification
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #1015, opened 2026-07-08
fix(cli): reject extra render_scene args

psiddharthdesign:
## Summary\n- make the render_scene MCP input parser strict so it matches its additionalProperties: false schema\n- add regression coverage for unknown render_scene arguments\n\n## Tests\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1017, opened 2026-07-08
chore(cli): name query scene record type

psiddharthdesign:
## Summary
- Add a local QuerySceneRecord alias for query-scene MCP argument and record helpers.
- Keep the change scoped to CLI MCP query scene typing only.

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1018, opened 2026-07-08
fix(cli): resolve query scene paths

psiddharthdesign:
## Summary
- Resolve trimmed query-scene MCP paths before stat/read/error reporting
- Add coverage for padded relative scene paths

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1025, opened 2026-07-08
Test env helper mutation restore

psiddharthdesign:
## Summary
- Add coverage that withEnvVar restores the original environment value even if the callback mutates it.

## Verification
- pnpm run build && node --test --test-concurrency=1 dist/testUtils/env.test.js
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #1034, opened 2026-07-08
Tighten create_scene JSON typing

psiddharthdesign:
## Summary\n- Keep parsed create_scene input as unknown until the runtime object guard passes\n- Cast to SceneJson only at the buildSceneBytes boundary\n\n## Verification\n- pnpm --dir cli test -- createScene: passed (414 tests)\n- pnpm typecheck: fails on existing app-wide TypeScript errors outside this change


psiddharthdesign:
@mergifyio queue

PR #1038, opened 2026-07-08
Tighten CLI scene validator set types

psiddharthdesign:
## Summary
- narrow CLI scene validation lookup sets to their literal union types
- keep runtime validation behavior unchanged while making future drift easier to catch in TypeScript

## Tests
- pnpm --dir cli test


psiddharthdesign:
@Mergifyio queue

PR #1043, opened 2026-07-08
Tighten render command dependency typing

psiddharthdesign:
## Summary
- Mark render command dependency injection hooks as readonly
- Wrap the mkdirSync dependency type for readability

## Verification
- pnpm --dir cli test -- render.test.ts

Note: root `pnpm typecheck` currently fails on existing app-wide TypeScript errors outside this CLI type-only change.


psiddharthdesign:
@mergifyio queue

PR #1048, opened 2026-07-08
test(cli): cover env cleanup after callback failures

psiddharthdesign:
## Summary
- add regression coverage for withEnvVar cleanup when the variable starts unset and the callback throws

## Tests
- pnpm --dir cli run build && node --test cli/dist/testUtils/env.test.js
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #1050, opened 2026-07-08
Stabilize driver error sentinel test

psiddharthdesign:
## Summary
- simplify the plain-text error sentinel fake app fixture to import fs the same way as the other driver fixtures
- avoids the fixture occasionally failing to write its sentinel before the render timeout

## Tests
- pnpm --dir cli run build && node --test --test-concurrency=1 cli/dist/electron/driver.test.js
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1063, opened 2026-07-09
test(cli): cover app bundle override slash

psiddharthdesign:
## Summary
- add CLI locator coverage for macOS .app bundle overrides with a trailing slash

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1065, opened 2026-07-09
test(cli): cover parented camera validation

psiddharthdesign:
## Summary
- add CLI scene validation coverage for cameras parented inside frames
- documents the existing single scene-level camera constraint through a regression test

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1071, opened 2026-07-09
Clarify malformed query scene fixture

psiddharthdesign:
## Summary\n- adds a named helper for the malformed query-scene children fixture\n- documents that the fixture represents stale non-string child entries\n\n## Tests\n- pnpm --dir cli test


psiddharthdesign:
@Mergifyio queue

PR #1072, opened 2026-07-09
test(cli): cover extension-only render suffixes

psiddharthdesign:
## Summary\n- add render format inference coverage for extension-only output names with query/hash suffixes\n\n## Verification\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1074, opened 2026-07-09
Tighten render option type guards

psiddharthdesign:
## Summary
- widen internal render option lookup sets to string sets
- remove assertions from render format and quality type guards

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1083, opened 2026-07-09
chore(cli): name scene metadata json type

psiddharthdesign:
## Summary
- add an explicit exported SceneMetaJson type for CLI scene authoring metadata
- update command tests to use the named scene metadata/scene JSON types directly

## Verification
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1085, opened 2026-07-09
test(cli): cover extension-only render outputs

psiddharthdesign:
## Summary
- add command-level coverage for extension-only render output filenames like .WEBM
- verifies the inferred render format is forwarded to the render driver

## Tests
- pnpm --dir cli test -- --test-name-pattern "render command"
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #1086, opened 2026-07-09
Avoid camera warning for camera-free scenes

psiddharthdesign:
## Summary
- suppress the active-camera warning when a scene has no camera nodes
- keep the warning for scenes that include cameras but have no active camera
- update CLI and MCP validation expectations

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1088, opened 2026-07-09
test(cli): cover render driver failures

psiddharthdesign:
## Summary
- Add CLI render command regression coverage for desktop driver failures.
- Verify the command reports the driver error and exits with code 1.

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #1090, opened 2026-07-09
Tighten render option guard types

psiddharthdesign:
## Summary\n- type render option lookup sets as string sets\n- remove unnecessary casts from render format and quality guards\n\n## Tests\n- pnpm test


psiddharthdesign:
@mergifyio queue

PR #1092, opened 2026-07-09
test(cli): keep create scene output docs in sync

psiddharthdesign:
## Summary
- Reuse one shared create_scene output-path description across Zod and MCP JSON schemas
- Update the schema assertion so the exposed MCP contract includes the parent-directory behavior

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1097, opened 2026-07-09
Tighten query scene schema test helper

psiddharthdesign:
## Summary
- add a small helper for query-scene MCP schema property assertions
- keep schema field checks the same while pairing the property cast with a runtime shape assertion

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #1119, opened 2026-07-09
fix(cli): clarify missing query scene files

psiddharthdesign:
## Summary
- report missing query MCP scene files with a clear file-not-found message
- update the query scene MCP test expectation for the clearer error

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #1126, opened 2026-07-09
chore(cli): reuse MCP args type for query tools

psiddharthdesign:
## Summary\n- Reuse the shared MCP tool args type in query scene MCP handlers\n- Remove the local duplicate Record alias\n\n## Verification\n- pnpm --dir cli test\n\nNote: root pnpm typecheck currently fails on unrelated existing app TypeScript errors outside this CLI change.


psiddharthdesign:
@mergifyio queue

PR #1134, opened 2026-07-09
chore(cli): type query scene schema helper

psiddharthdesign:
## Summary
- type the query scene schema test helper against the shared MCP Tool shape
- keep the change scoped to CLI test code

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1136, opened 2026-07-09
Mark CLI command options readonly

psiddharthdesign:
## Summary
- Mark CLI create/render command option interfaces as readonly to reflect read-only handler usage.

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1139, opened 2026-07-09
chore: ignore nested dist outputs in lint

psiddharthdesign:
## Summary\n- Extend ESLint generated-output ignores to cover package-level dist directories such as cli/dist.\n\n## Verification\n- pnpm exec eslint cli/dist/mcp/tools/schema.d.ts --no-warn-ignored\n- pnpm test\n\nNote: top-level pnpm lint and pnpm typecheck still have unrelated pre-existing failures on main; this change verifies the generated-file lint case directly.


psiddharthdesign:
@mergifyio queue

PR #1143, opened 2026-07-09
test(cli): cover async env helper failures

psiddharthdesign:
## Summary
- add a focused regression test for withEnvVar restoring existing env values after async callback failures

## Tests
- pnpm --dir cli run build && node --test cli/dist/testUtils/env.test.js
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1145, opened 2026-07-09
Ignore stale query scene track entries

psiddharthdesign:
## Summary\n- Ignore non-object persisted track map entries in the MCP list_tracks handler\n- Add a regression fixture that injects a stale scalar track entry into saved .hype bytes\n\n## Tests\n- pnpm --dir cli run build && node --test --test-concurrency=1 cli/dist/mcp/queryScene.test.js\n- pnpm test


psiddharthdesign:
@mergifyio queue

PR #1148, opened 2026-07-09
chore(cli): mark MCP tool args readonly

psiddharthdesign:
## Summary
- mark the shared MCP tool args type as readonly to match handler usage

## Verification
- pnpm --dir cli test

Note: root pnpm typecheck still fails on existing app-level TypeScript errors outside this CLI-only change.


psiddharthdesign:
@mergifyio queue

PR #1153, opened 2026-07-09
Guard get_layer against mismatched node ids

psiddharthdesign:
## Summary
- require get_layer results to match the requested node id exactly
- add a regression test for stale/malformed node map entries

## Tests
- pnpm --dir cli run build && node --test --test-concurrency=1 cli/dist/mcp/queryScene.test.js
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1158, opened 2026-07-09
chore(cli): resolve info scene paths

psiddharthdesign:
## Summary
- Resolve MCP info_scene inputs before reading so relative paths are normalized like the CLI info/open paths.
- Add regression coverage for relative scene paths.

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1160, opened 2026-07-09
chore(cli): ignore malformed query scene nodes

psiddharthdesign:
## Summary
- skip malformed non-object node entries when listing scene layers
- extend the query-scene MCP test fixture to cover stale node entries alongside stale tracks

## Tests
- pnpm --dir cli test -- queryScene


psiddharthdesign:
@mergifyio queue

PR #1162, opened 2026-07-09
chore(cli): guard empty MCP args helper

psiddharthdesign:
## Summary
- Return a clean MCP validation error when the shared empty-args helper receives non-object arguments.
- Add focused coverage for array and null malformed arguments.

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1163, opened 2026-07-09
fix(cli): reject directory render outputs

psiddharthdesign:
## Summary
- reject `render_scene` requests when the resolved output path already exists as a directory
- add a regression test so the MCP tool returns a clear error before locating the desktop app

## Tests
- `pnpm --dir cli test`


psiddharthdesign:
@mergifyio queue

PR #1164, opened 2026-07-09
Harden empty MCP arg validation

psiddharthdesign:
## Summary\n- Tighten empty-argument MCP validation to accept only plain JSON-style records\n- Add coverage for null-prototype records and non-plain object rejection\n\n## Tests\n- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1166, opened 2026-07-09
Validate create output before building

psiddharthdesign:
## Summary
- validate the create command output path before building scene bytes
- add regression coverage for directory outputs masking scene-build failures

## Tests
- pnpm --dir cli test -- create.test.js
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #1167, opened 2026-07-09
chore(cli): preserve readonly partial arrays

psiddharthdesign:
## Summary
- Preserve readonly array properties in the CLI scene builder DeepPartial helper.
- Keep the change type-only and scoped to scene authoring types.

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1168, opened 2026-07-09
fix(cli): resolve validate scene paths

psiddharthdesign:
## Summary
- Resolve MCP validate_scene inputs before stat/read, matching other scene MCP tools
- Add a regression test for relative validate_scene paths

## Tests
- pnpm test (in cli/)


psiddharthdesign:
@mergifyio queue

PR #1170, opened 2026-07-09
test: cover patched node parent linking

psiddharthdesign:
## Summary
- Add a focused CLI scene patch regression test for `createNode` parent-child wiring.
- Validate the patched scene after linking the new node into the root children list.

## Tests
- `pnpm --dir cli test`


psiddharthdesign:
@mergifyio queue

PR #1176, opened 2026-07-10
Reject directory render outputs

psiddharthdesign:
## Summary
- Reject directory paths passed to the CLI render output option before locating or launching the desktop app
- Add a regression test covering the preflight failure

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1177, opened 2026-07-10
Guard malformed camera query results

psiddharthdesign:
## Summary
- add a typed camera-node guard for the CLI MCP list_cameras handler
- cover malformed camera-shaped persisted nodes with a regression test

## Verification
- pnpm --dir cli test

Note: repo-wide pnpm typecheck still fails on existing app-side TypeScript errors outside this change.


psiddharthdesign:
@mergifyio queue

PR #1179, opened 2026-07-10
chore(cli): reuse MCP string schema type

psiddharthdesign:
## Summary
- Reuse the shared MCP string schema property type in query scene tools
- Remove the local duplicate type definition

## Tests
- pnpm test


psiddharthdesign:
@mergifyio queue

PR #1182, opened 2026-07-10
test(cli): cover null-prototype MCP arg keys

psiddharthdesign:
## Summary
- add coverage for unexpected keys on null-prototype MCP argument records
- keep empty-args schema helper behavior deterministic for agent tools

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1188, opened 2026-07-10
test(cli): cover output path stat failures

psiddharthdesign:
## Summary
- add a focused render command regression test for existing output file stat failures
- verify the command exits before locating the desktop app or invoking the render driver

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1190, opened 2026-07-10
Tighten render command option types

psiddharthdesign:
## Summary
- Mark defaulted render command options as optional in the local Commander option type
- Keep existing runtime defaults and validation unchanged

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1191, opened 2026-07-10
test(cli): cover env var deletion restores

psiddharthdesign:
## Summary
- add coverage that withEnvVar restores an originally set env var when the callback deletes it

## Tests
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1203, opened 2026-07-10
Bound headless render error sentinel waits

psiddharthdesign:
## Summary
- Preflight the headless render app path for executability after stale output/sentinel cleanup.
- Fall back to a generic render error when an error sentinel remains unreadable after the spawned process exits and the grace window elapses.
- Cover doctor report render formats and quality presets against the shared constants.

## Tests
- pnpm --dir cli run build && node --test --test-concurrency=1 cli/dist/electron/driver.test.js
- pnpm --dir cli test


psiddharthdesign:
@mergifyio queue

PR #1210, opened 2026-07-10
Validate CLI fps as decimal input

psiddharthdesign:
## Summary
- require CLI render --fps values to be plain decimal integer input
- add a regression test for JavaScript numeric notation like 1e2 and 0x10

## Verification
- pnpm --dir cli test
- pnpm exec eslint cli/src/commands/render.ts cli/src/commands/render.test.ts

## Notes
- pnpm typecheck currently fails on unrelated app-level TypeScript errors outside this change.
- pnpm lint currently fails on unrelated app-level ESLint errors outside this change.


psiddharthdesign:
@mergifyio queue

PR #1223, opened 2026-07-10
test(cli): reduce render driver sentinel flakes

psiddharthdesign:
## Summary\n- make the malformed success sentinel fixture complete on the next event-loop turn instead of a wall-clock timer\n- keep the trimmed JSON error sentinel visible briefly so the polling loop reliably observes it under load\n\n## Verification\n- pnpm run build && node --test --test-concurrency=1 --test-name-pattern='driveHeadlessRender (ignores malformed success sentinels|trims JSON error sentinel messages)' dist/electron/driver.test.js\n\nNote: a full `pnpm test` run was attempted before this narrower fix and exposed broader pre-existing render-driver timing flakes in other cases under heavy load.


psiddharthdesign:
@mergifyio queue

PR #1229, opened 2026-07-11
Stabilize render scene path normalization test

psiddharthdesign:
## Summary
- avoid launching a fake child process in the padded scene-path MCP render test
- assert the normalized scene path through the injected render dependency instead

## Tests
- pnpm --dir cli run build && node --test --test-concurrency=1 cli/dist/mcp/renderScene.test.js


psiddharthdesign:
@mergifyio queue

PR #1246, opened 2026-07-17
Mark query scene snapshot readonly

psiddharthdesign:
## What changed

- Mark the internal query-scene snapshot fields as `readonly`.

## Why

The snapshot represents parsed scene data that query handlers only inspect. Encoding that immutability in the type prevents accidental assignments without changing runtime behavior or the public API.

## Impact

No user-visible behavior changes. This is an internal type-safety maintenance improvement.

## Validation

- `pnpm --dir cli test` (457 tests passed)
- `pnpm exec eslint cli/src/mcp/tools/queryScene.ts`
- `git diff --check`


psiddharthdesign:
@mergifyio queue

PR #1253, opened 2026-07-20
Mark capability payloads readonly

psiddharthdesign:
## What changed

Marked the internal MCP capability response payload fields as `readonly`, including the nested validation flag.

## Why

These payloads are assembled once and immediately serialized. Encoding that immutability in their TypeScript types prevents accidental mutation while keeping runtime behavior unchanged.

## Impact

No user-facing or runtime behavior changes.

## Validation

- `pnpm --dir cli test` (461 passed)
- `pnpm exec eslint cli/src/mcp/tools/capabilities.ts`
- `git diff --check`


psiddharthdesign:
@mergifyio queue

PR #1274, opened 2026-07-27
Tighten stagger text test types

psiddharthdesign:
## What changed

- Preserve the guaranteed textAnimation type in the text-track test fixture.
- Reuse narrowed text nodes in assertions so TypeScript retains the node kind.

## Why

The fixture always provides text animation metadata, but its broad return type discarded that guarantee and repeated node lookups prevented type narrowing.

## Impact

Test-only type safety improvement; production behavior is unchanged.

## Validation

- pnpm exec vitest run src/anim/staggerSets.test.ts (27 tests passed)
- pnpm exec eslint src/anim/staggerSets.test.ts

### DIFF
--- a/src/anim/staggerSets.test.ts
+++ b/src/anim/staggerSets.test.ts
@@ -2,7 +2,7 @@
 
 import { describe, expect, it } from 'vitest'
 import * as Y from 'yjs'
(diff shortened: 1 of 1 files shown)


psiddharthdesign:
@mergifyio queue

PR #1275, opened 2026-07-28
Fix stagger test type import

psiddharthdesign:
## What changed

- Import TextAnimationConfig from its canonical textAnimations module instead of the scene barrel.

## Why

The scene barrel does not export this type, so the test file produced a TypeScript error even though the runtime tests passed.

## Impact

Test-only type-safety maintenance; no production behavior changes.

## Validation

- pnpm exec vitest run src/anim/staggerSets.test.ts — 27 tests passed
- pnpm exec eslint src/anim/staggerSets.test.ts — passed
- Repository typecheck no longer reports staggerSets.test.ts; unrelated pre-existing type errors remain elsewhere.

### DIFF
--- a/src/anim/staggerSets.test.ts
+++ b/src/anim/staggerSets.test.ts
@@ -2,11 +2,14 @@
 
 import { describe, expect, it } from 'vitest'
 import * as Y from 'yjs'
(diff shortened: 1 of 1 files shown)


psiddharthdesign:
@mergifyio queue

PR #1289, opened 2026-08-23
Clarify capture frame source roles

psiddharthdesign:
## Summary
- clarify how the hidden render-window pipeline uses the capturePage frame source
- document the editor instance as the legacy/native fallback
- update the comparison text to reflect the shared DOM scene painter

## Verification
- `pnpm exec eslint src/export/captureRect.ts`
- `git diff --check main...HEAD`

### DIFF
--- a/src/export/captureRect.ts
+++ b/src/export/captureRect.ts
@@ -3,11 +3,14 @@
 /**
  * Frame source built on Electron's `webContents.capturePage`.
  *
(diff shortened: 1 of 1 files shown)


psiddharthdesign:
@mergifyio queue

psiddharthdesign:
@mergifyio queue

PR #1291, opened 2026-08-23
Clarify tab capture MP4 behavior

psiddharthdesign:
Corrects stale tab-capture documentation: MP4 requests routed here now surface an error, while native WebCodecs is the supported MP4 path.

Verification:
- focused source-consistency assertion
- git diff --check
- exact one-file scope; no manifest or lockfile changes

Full ESLint installation was unavailable because required cached tarballs were missing.


psiddharthdesign:
@mergifyio queue

